PR #12, opened 2017-01-03
feat(*): add demo mode

btry:
Self registering an account sends a welcome email with a validation hash
Valdiation of a self registered account must be achieved in a reasonable of a few days (cron task)
Self registered account created with demo mode enabled will be disabled after a a trial period (cron task)


btry:
done

PR #13, opened 2017-01-05
chore(ci): run tests with php 7.1 and GLPi with backports

btry:
Flyve MDM needs features in GLPi not yet released


btry:
replaced by #16

PR #50, opened 2017-06-08
Agent feedback

btry:
Handles agent online status
sends a task ID to agents with policies
expects agents send feedback about policies they shall apply. 

BREAKING CHANGE:
Fleet_Policy renamed into Task


btry:
rebased and reworked to be based on atou unit tests (see#100)

This one is now obsolete.

PR #82, opened 2017-07-26
Maintain online status of devices

btry:
Feature extracted from #50 

use will and testament to maintain online status of devices


install/mysql/plugin_flyvemdm_empty.sql
```diff
@@ -16,6 +16,7 @@ CREATE TABLE IF NOT EXISTS `glpi_plugin_flyvemdm_agents` (
   `plugin_flyvemdm_fleets_id` int(11)                                           DEFAULT NULL,
   `last_report`               datetime                                          DEFAULT NULL,
   `last_contact`              datetime                                          DEFAULT NULL,
+  `online`                    tinyint(1)                                        NOT NULL DEFAULT '0',
```
ajsb85: String or int?
Check please

btry: this is a boolean, and in GLPI a bool is encoded with a 1 or a 0 in a single digit column

PR #107, opened 2017-08-16
WIP Daemon reimplementation

btry:
Still needs an enhancement to properly stop the daemon

To get rid of problems with the hardcoded user in old service.sh, this daemon should be started by a cron like this

```
@reboot service.php
```

with crontab -e www-data (or the user runnung Apache / Nginx)


ajsb85:
https://www.phpclasses.org/blog/post/270-Creating-a-PHP-Daemon-Service.html

ajsb85 (dismissed):
Reviewed

ajsb85 (dismissed):
Reviewed

btry:
Still need to fix the signal handlers. They dont work with the current code

DIOHz0r:
Hello, Is this issue still WIP?

btry:
Yes, the daemon fails to catch signals when trying to stop it. 

This is low priority as the current implementation works.

ingluife:
Hi, @flyve-mdm/backend-development.

What will we do with this?

I can help with the conflict

btry:
The conflict is not the problem. With this PR the daemon cannot be catch the signals yet. A rebase is necessary, but the branch is still unfinished

PR #109, opened 2017-08-16
[WIP] retain type of mdm for agents

btry:
also cleanup dead code, refactor installer, refactor old fleet_policy itemtype clues into task

agent have a new property : **mdm_type*. Dashboards can use it to determine the t ype of a device. **android* and **apple** supported, but communication layer for apple not implemented yet.

when a new layer of communication will be added, the mdm_type will handle other values.


inc/agent.class.php
```diff
@@ -67,6 +67,16 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
    protected $topic = null;
 
    /**
+    * get mdm types availables
+    */
+   public static function getEnumMdmType() {
+      return array(
+         'android'   => __("Flyve MDM"),
```
ajsb85: flyve is a type?

ajsb85: Why not Android instead Flyve MDM?

btry: I began with a type named flyve, and change to android without updating the label. fixing...

PR #117, opened 2017-08-30
Fix user deletion

### LINKED ISSUE #98: Don't delete user when we delete his last device
When we delete an agent the backend deletes the associated computer and if the associated user no longer has any computer, he is deleted too.

We should keep the user.

DIOHz0r:
Hi, @btry.

This Pull-Request fix user deletion when no computers left.

Close #98


tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -644,19 +648,22 @@ public function testPurgeEnroledAgent() {
...
+
+       $this->boolean($user->getFromDb($testUserId))->isTrue();
    }
 
    /**
-    * Test the purge of an agent deletes the user if he no longer has any agent
-    *
+    * Test the purge of an agent the user must persist if he no longer has any agent
     *
+    * @tags thisIsOne
```
btry: What is the purpose of this annotation ?

inc/agent.class.php
```diff
@@ -554,12 +554,12 @@ public function pre_deleteItem() {
...
 
          // Check the user still has one or several profiles
-         $rows = $profile_User->find("`users_id`='$ownerUserId'", '', '1');
+         /*$rows = $profile_User->find("`users_id`='$ownerUserId'", '', '1');
          if (count($rows) == 0) {
             // Delete the user
             $user = new User();
             $user->delete(['id' => $ownerUserId], true);
-         }
+         }*/
```
btry: OK for me then you can drop the disabled code

tests/suite-integration/PluginFlyvemdmAgent.php
DIOHz0r: atoum's tags is a helper to do only that method when you run the test if you use the -t argument. http://docs.atoum.org/en/latest/running_tests.html#filtres-par-tag

btry:
history rewritten in #119 (and merged)

Issue #126, opened 2017-09-28
Refactor package upload code


btry:
In GLPI exists a function isAPI (autoload.function.php)

It is way better to use it to improve prepareInputForAdd() and prepareInputForUpdate() in PluginFlyvemdmPackage to detect if the creation or update of a package is done from the API or the UI.

By the way, these methods need refactor, there are issues. Can be fixed at the same time

DIOHz0r:
Changes merged

PR #131, opened 2017-10-04
Refactor package upload code

DIOHz0r:
Hi, @btry 

This is a solution for refactor the package upload code from the issue #126 

Best regards.

### DIFF
--- a/inc/package.class.php
+++ b/inc/package.class.php
@@ -55,7 +55,8 @@ class PluginFlyvemdmPackage extends CommonDBTM {
 
    /**
     * Localized name of the type
-    * @param $nb  integer  number of item in the type (default 0)
+    * @param integer $nb number of item in the type (default 0)
+    * @return string
     */
    public static function getTypeName($nb=0) {
       return _n('Package', 'Packages', $nb, "flyvemdm");
@@ -85,8 +86,7 @@ class PluginFlyvemdmPackage extends CommonDBTM {
     * Returns the tab name of this itemtype, depending on the itemtype on which it will be displayed
     * If the tab shall not display then returns an empty string
     * @param CommonGLPI $item on which the tab will show
-    * @param number $withtemplate template mode for $item : 0 = no template - 1 = edit template - 2 = from template
-    *
+    * @param int $withtemplate template mode for $item : 0 = no template - 1 = edit template - 2 = from template
     * @return string
     */
    public function getTabNameForItem(CommonGLPI $item, $withtemplate=0) {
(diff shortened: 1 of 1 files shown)


inc/package.class.php
```diff
@@ -148,10 +148,13 @@ protected static function getMaxFileSize() {
       switch ($last) {
          case 'g':
             $val *= 1024;
+            break;
```
btry: no break: this is wanted

inc/package.class.php
```diff
@@ -148,10 +148,13 @@ protected static function getMaxFileSize() {
...
          case 'm':
             $val *= 1024;
+            break;
```
btry: neither here

inc/package.class.php
```diff
@@ -148,10 +148,13 @@ protected static function getMaxFileSize() {
...
          case 'k':
             $val *= 1024;
+            break;
```
btry: and here

If we have 256g as argument 256 will be multiplied 3 times by 1024 to be converted into bytes.

inc/package.class.php
```diff
@@ -602,4 +604,27 @@ public function hook_entity_purge(CommonDBTM $item) {
...
    }
+
+   /**
+    * @param array $input
+    * @param \ApkParser\Parser $apk
+    * @param string $destination
+    * @param string $uploadedFile
+    * @return array
+    */
+   protected function setApkManifest(array $input, \ApkParser\Parser $apk, $destination, $uploadedFile)
```
btry: I would name this method soemthing like getApkMetadata() or getApkManifest() because it gets metadata,  Set seems not appropriate here.

Also, if there is no identified need to inherit and share the class, it is better to declare it private first. If we need ti make it protected it is less risky  than making it protected and later change it to private.

inc/package.class.php
DIOHz0r: ok

inc/package.class.php
DIOHz0r: ok

inc/package.class.php
DIOHz0r: ok

inc/package.class.php
DIOHz0r: I have the doubt about this because my IDE show this alert "Case clause execution can unintentionally falls through the next case, which is most often an error. A break or a return statement at the end of the case will prevent this error."

ajsb85:
Hi, again @DIOHz0r 

Remember to put this PR in a proper kanban list of the project board.

Thanks.

btry:
hi

 The PR looks good. In phpdoc I prefer **integer** instead of **int**, but I won't request change for such such things.

In the history of this branch I see some commits from an older and already merged PR : #119 which is a history rewrite of #117. 

I guess this branch is not based on develop from this repo, but your older contributions in your local develop branch. Can you solve this lcoally, then force push the result here ? The result should be the same history without the first commits
image

Also, all expected commits may be merged into a signe one named **refactor(package): <your favourite description>** before merge.

DIOHz0r:
Hi

Yes I notice that olds commits so I'm going to fix that history and rebase some changes.

DIOHz0r:
Hi @btry, 

Please review this PR

ajsb85:
Hi, @DIOHz0r 
@btry scheduled the review of this PR for the next week 👍 
Good job.

Issue #32, opened 2017-05-02
Enroll device once delete it from website


rafaelje:
### Expected behaviour
Enroll my device once delete it from website

### Actual behaviour
When I delete some device connected from website is impossible to reconnect this device again because I can't read the QR or logout

### Environment Details

- Flyve MDM version: 0.99.46

### Steps to reproduce the behaviour

1. Enroll a device
2. Create a fleet
3. Send fleet
4. Delete the device on website
5. Check the device
6. Try to Enroll again

btry:
A QR code can ne used only once (except if a debug setting is enabled).

We still need to choose if we must try to unenroll a device if the admin deletes it. Currently, you need to stop the agent, reset its data and restart it to be able to enroll again. 

To properly reset an agent, unenroll it instead of deleting it. The agent will receive an unenrollment message and reset itself.

PR #165, opened 2017-11-06
Improve device enrollment and deletion

DIOHz0r:
Hello,

This pull request is for resolve the issue #32 allowing devices to enroll again once they are deleted and this will solve issue #172 about enrollment process.

Regards.


inc/agent.class.php
```diff
@@ -588,22 +588,27 @@ public function pre_deleteItem() {
...
       ], true);
 
+      // Send unrolling message to reset agent status
+      if ($this->fields['enroll_status'] == 'enrolled') {
+         $this->sendUnenrollQuery();
+      }
+
       // Delete the MQTT user for the agent
-      if (!empty($serial)) {
+      /*if (!empty($serial)) {
```
btry: I think we shall delete the MQTT account because the agent is deleted.

inc/agent.class.php
```diff
@@ -1091,12 +1096,17 @@ protected function enrollByInvitationToken($input) {
...
-      if (count($rows) > 0) {
-         $event = __('The serial already exists', 'flyvemdm');
-         $this->filterMessages($event);
-         $this->logInvitationEvent($invitation, $event);
-         return false;
+      $computers = $computer->find("`entities_id`='$entityId' AND `serial`='$serial'");
+      if (count($computers) > 0) {
+         // A computer with the same serial exist, let's find if an agent exist for that device.
+         $flyvemdmAgent = new PluginFlyvemdmAgent();
+         $agents = $flyvemdmAgent->find("`entities_id`='$entityId' AND `computers_id`='{$computer->getID()}'");
```
btry: {$computer->getID()} <=== the syntax {$obj->method()} is too messy. I mrefer using a var to store the value of the method then use it here.

inc/agent.class.php
```diff
@@ -1167,20 +1177,32 @@ protected function enrollByInvitationToken($input) {
...
-         'serial'       => $serial,
-         'uuid'         => $uuid,
-      ]);
-      if ($computerId === false) {
-         $event = __("Cannot create the device", 'flyvemdm');
-         $this->filterMessages($event);
-         $this->logInvitationEvent($invitation, $event);
-         return false;
+      // Create/update the device
+      if (count($computers) > 0) {
```
btry: what if count of computers > 1 ?

Which one will you choose ?

btry:
Hi

Globally this workflow may fail. I'm more and more in favor of an import of the computer from the inventory XML. 

I suggest isntead of patching the enrollemnt and deletion workflows we use the import rules engine. This way, the admin of Flyve MDM may choose what happens and use more sophisticated rules to match a enrolling device to an existing computer.

inc/agent.class.php
DIOHz0r: ok

inc/agent.class.php
DIOHz0r: ok

inc/agent.class.php
btry: When we unenroll a device the deletion of the agent occurs when the ACK is received.

see https://github.com/flyve-mdm/flyve-mdm-glpi-plugin/blob/develop/inc/mqtthandler.class.php#L124

DIOHz0r:
Hello,

Travis is passing the test, however there's an [error](https://travis-ci.org/flyve-mdm/glpi-plugin/jobs/319848860#L1024) with the test of Agent integration. 

One of the test on the Agent is making the rest of the tests unchecked and Travis is not taking this as an invalid test.

I've run the test locally and one test is giving and Exception and other one is triggering the empty log problem and stoping the rest of the tests.

I'll be checking and debuging this problem.

DIOHz0r:
Hello @btry,

I tested this change locally with one empty DB and it success so please accept it to merge.

Issue #176, opened 2017-11-14
Filse size option for the Item Type PluginFlyvemdmFile


ajsb85:
Hi, @btry and @DIOHz0r 

Could you add a key for the size of the file?

Thank you.

ajsb85:
Fyi, the `PluginFlyvemdmPackage`.

Use the key: `filesize`

btry:
Can you check the HTTP HEAD verb works first, when you try to download the file ?

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html (section 9.4)

I think this is better than maitnaining a value in the DB. This is a metadata of the file itself.

ajsb85:
No, it's not to download the file, this is to get all files, the list don't have the key filesize. But in the app list appear.

ajsb85:
Example of the package

```JSON
[
    {
        "id": 2,
        "name": "com.uberspot.a2048",
        "alias": 2048,
        "version": 2,
        "version_code": 20,
        "icon": "iVBORUVORK5CYII=",
        "filename": "/var/www/public/glpi/files/var/www/public/glpi/filesflyvemdm/package/0/59f33f8f84a16_59f33f83a0e786.9244290156fa7aa5ed741_com.uberspot.a2048_20.apk",
        "filesize": 633567,
        "entities_id": 0,
        "dl_filename": "59f33f83a0e786.9244290156fa7aa5ed741_com.uberspot.a2048_20.apk",
        "plugin_orion_tasks_id": 0,
        "parse_status": "parsed",
        "links": [
            {
                "rel": "Entity",
                "href": "https://dev.flyve.org/glpi/apirest.php/Entity/0"
            }
        ]
    },
```

ajsb85:
You can see the `filesize`

But take a look in the file:

```JSON
    {
        "id": 8,
        "name": "Icon.png",
        "source": "14/597babff83c94_Icon.png",
        "entities_id": 14,
        "version": 1,
        "links": [
            {
                "rel": "Entity",
                "href": "https://dev.flyve.org/glpi/apirest.php/Entity/14"
            }
        ]
    },
```

ajsb85:
Example of usage

![files-100](https://user-images.githubusercontent.com/663460/32791903-57077a94-c962-11e7-9409-7b95fc2cce8d.jpg)

btry:
I'll add he file size in a new DB column.

ajsb85:
@btry Thank you

[...]

PR #180, opened 2017-11-14
Update search options

### LINKED ISSUE #179: All keys of the Get Item Type Agent in the Search Option List
Hi, @btry 

Could you add the same data of the item type PluginFlyvemdmAgent in the listSearchOption endpoint?

Thanks.

btry:
Fix #179 

Waiting for possible other updates related to search options (#176 ?)


ajsb85:
Hi, @btry 

Could you add also `PluginFlyvemdmAgent.Computer.User.firstname`?

Thank you.

btry:
I think this is a bad idea to make available in an itemtype a data coming from an other itemtype which is not related to the first one. An agent and a user are not tightly coupled in a semantic point of view.

By the way we my need to remove the user email search option from the agent because we need to allow agents without any relatin to a (human) user in the future.

ajsb85:
@btry Actually I can get  PluginFlyvemdmAgent.Computer.User.realname in the list of the search options

ajsb85:
Actual search options in the Agent

```JSON
"PluginFlyvemdmAgent.name": "a.salas@ieee.org",
"PluginFlyvemdmAgent.id": 115,
"PluginFlyvemdmAgent.PluginFlyvemdmFleet.name": "not managed fleet",
"PluginFlyvemdmAgent.Computer.id": 153,
"PluginFlyvemdmAgent.Computer.serial": "LGK220AISOLZ55",
"PluginFlyvemdmAgent.Computer.User.id": 221,
"PluginFlyvemdmAgent.PluginFlyvemdmFleet.id": 7,
"PluginFlyvemdmAgent.last_contact": "2017-11-13 19:00:57",
"PluginFlyvemdmAgent.Computer.User.realname": "Salas Bastidas",
"PluginFlyvemdmAgent.version": "3.0.0",
"PluginFlyvemdmAgent.is_online": 1
```

btry:
hummm I forgot about this. There is a background about existence of this data, and I think this should be removed mostly because the design of the plugin is going to be better shortly.

ajsb85:
I will merge it @btry 
No continuous deployment at the moment?

btry:
Hi

@ajsb85 I think I should remove the serial number from search options. We can get the ID of the computer, then we can get the serial number indirectly, the same way we retrieve the user name.

PR #181, opened 2017-11-14
Set unique agent account name.

### LINKED ISSUE #177: Agent username not unique
### Expected behaviour
The GLPI user account of the agent registered must be unique

### Actual behaviour
When a new agent is trying to do the enroll process a user account is created using the serial number of the device, if  the username is already registered the agent fails to do the enrollment.

### Environment Details

- GLPi version: 9.2
- Flyve MDM version: 1.0
- FusionInventory version:  9.2
- Mosquitto version: 3.1
- Mosquitto's auth plugin  version: 0.1.2

You have multiple ways to do this: 
Case 1:
1. Create one User in GLPI using the device serial
2. Try to enroll the device.

Case 2:
1. Enroll The device
2. Delete & purge the computer and the agent at GLPI leaving the User created, all this without the device connected
3. Try to enroll again the device.

Case 3 (most dificult but almost like case 1):
1. Enroll a device 
2. Try to enroll other device with the same serial.

imagen

DIOHz0r:
## Checklist

- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

When a device is trying to enroll the username could exist because its login is the serial number from the device.

Fixes #177 

## What is the new behavior?
The agent user account now has a unique id. 

- [ ] Yes
- [x] No

### DIFF
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -1116,7 +1116,8 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
       $agentAccount = new User();
       $agentAccount->add([
          'usercategories_id' => $config['agentusercategories_id'],
-         'name'              => $serial,
+         'name'              => 'flyvemdm-' . PluginFlyvemdmCommon::generateUUID(),
+         'realname'          => $serial,
          '_profiles_id'      => $config['agent_profiles_id'],
          'profiles_id'       => $config['agent_profiles_id'],      // Default profile when user logs in
          '_entities_id'      => $entityId,
--- a/inc/common.class.php
+++ b/inc/common.class.php
@@ -91,4 +91,29 @@ class PluginFlyvemdmCommon
       return $enum;
    }
 
+   /**
+    * @return string pseudo-random UUID version 4
+    */
+   public static function generateUUID() {
+      return sprintf('%04x%04x-%04x-%04x-%04x-%04x%04x%04x',
+
+         // 32 bits for "time_low"
+         mt_rand(0, 0xffff), mt_rand(0, 0xffff),
+
+         // 16 bits for "time_mid"
+         mt_rand(0, 0xffff),
+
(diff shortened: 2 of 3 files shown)


inc/agent.class.php
```diff
@@ -1116,7 +1116,8 @@ protected function enrollByInvitationToken($input) {
       $agentAccount = new User();
       $agentAccount->add([
          'usercategories_id' => $config['agentusercategories_id'],
-         'name'              => $serial,
+         'name'              => 'flyvemdm-' . uniqid(),
```
btry: OK for the idea, but we need to check that uniqid() is "strong" enough to avoid collisions over time. 

Also this has impact on the credentials used by he agent to consume  the REST API. As the login is not predictible from the point of view of the agent, we need to enhance the invitation process because the agents use serial as login and a password provided in the invitation data to consule the REST API and enroll themselves.

DIOHz0r: OK, I will add some logic to be more unique. I have talk with @rafaelje about this and the other part of the solution would be to return the agent account name to the device and store it to other uses.

btry: Have a look on this: https://stackoverflow.com/questions/1155008/how-unique-is-uuid

Also, my badn the enrollment only uses a user token so this PR will be kelt simple (but I still have some ideas :) )

btry: I think that using a UUID generator would be a good option. Here is one here http://php.net/manual/en/function.uniqid.php#94959

We might use UUID version 3 or 5 for username, and define their namespace with a V4 UUID.

btry:
Even better than a class in the project to implement UUID: a package

https://github.com/ramsey/uuid

DIOHz0r:
The UUID is a cool solution but IMO that would be too much code just for one single application. With the last change I did something more simple and probably more unique than the first commit.

PR #194, opened 2017-11-19
Fix misordered list

btry:
A unit test fails. Here is the fix. This is only a misordered vars in the list.


btry:
We need to see why an uncompleted test did not triggered a failure.
Here is the logs of the current develop branche
image

PR #197, opened 2017-11-20
[WIP] Tests additions and upgrades

DIOHz0r:
## Checklist

- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Unit test

There are a lot of missing unit tests

## What is the new behavior?
New unit test for the plugin

- [ ] Yes
- [x] No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
 
 before_script:
   - mysql -u root -e 'create database glpitest;'
-  - phpenv config-rm xdebug.ini || true
   - rm -f composer.lock
   - tests/config-composer.sh
   - git clone --depth=1 $GLPI_SOURCE ../glpi && cd ../glpi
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -18,10 +17,10 @@ before_script:
   - cd plugins/flyvemdm && composer install --no-interaction
 
 script:
-  - vendor/bin/atoum --debug -bf tests/bootstrap.php -no-cc --max-children-number 1 -d tests/suite-install
-  - vendor/bin/atoum --debug -bf tests/bootstrap.php -no-cc --max-children-number 1 -d tests/suite-unit
-  - for f in `ls tests/suite-integration/`; do vendor/bin/atoum --debug -bf tests/bootstrap.php -no-cc --max-children-number 1 -f tests/suite-integration/$f; done;
-  - vendor/bin/atoum --debug -bf tests/bootstrap.php -no-cc --max-children-number 1 -d tests/suite-uninstall
+  - vendor/bin/atoum -ft -bf tests/bootstrap.php --max-children-number 1 -d tests/suite-install -ncc
+  - for f in `ls tests/suite-unit/`; do vendor/bin/atoum -ft -bf tests/bootstrap.php --max-children-number 1 -f tests/suite-unit/$f; done;
```
btry: I disagree with the change on the line for suite-unit tests. Using a for loop introduces the risk that Travis will not detect failure for a file in the folder. This is the other line for suite-intrgration which should change to be similar to this one, but we need to fond why the other form fails in Travis and not locally (my computer).

I expect the issue will disappear when we will migrate and rewrite some or most tests in suite-integration

btry: I disagree with the change on the line for suite-unit tests. Using a for loop introduces the risk that Travis will not detect failure for a file in the folder. This is the other line for suite-intrgration which should change to be similar to this one, but we need to fond why the other form fails in Travis and not locally (my computer).

I expect the issue will disappear when we will migrate and rewrite some or most tests in suite-integration

DIOHz0r: There is a issue for the test with atoum in this project and this is to avoid the error, You can check it in atoum/atoum#780

btry: I think this is not the quantity of files which fails. In GLPI there is a folder with way more than 11 files and it succeeds with -d path/to/folder/ argument. 

I think there is a different failure but you did not find the thing which triggers it, like I encounteered in the integration suite. To workaround, I suggest we enhance /replace the for loop to ensure Travis will catch a possible error.

ajsb85:
Hi, @btry 
Any another review? 
Or we can merge it?

btry:
It is required to find an other workaround for the issue @DIOHz0r found, and by  the way apply it too to  the suite-integration tests. Merging as is will suceed tests no matter there are failures (except in isntall and uninstall tests). This is too dangerous.

To speedup things I suggest we add test files one by line (but its ugly)

DIOHz0r:
Ok, I will change the filter used for test-unit folder as it was before.

PR #220, opened 2017-11-23
Upgrade should not modify package file path more than once

btry:
After an upgrade packages could not be downloaded

### DIFF
--- a/install/installer.class.php
+++ b/install/installer.class.php
@@ -110,7 +110,6 @@ class PluginFlyvemdmInstaller {
             $this->migration->displayWarning("Error creating tables : " . $DB->error(), true);
             return false;
          }
-
          $this->createInitialConfig();
       } else {
          if ($this->endsWith(PLUGIN_FLYVEMDM_VERSION, "-dev") || (version_compare(self::getCurrentVersion(), PLUGIN_FLYVEMDM_VERSION) != 0)) {
@@ -509,7 +508,6 @@ Regards,
          if (function_exists($updateFunction)) {
             $this->migration->addNewMessageArea("Upgrade to $toVersion");
             $updateFunction($this->migration);
-            $this->migration->executeMigration();
             $this->migration->displayMessage('Done');
          }
       }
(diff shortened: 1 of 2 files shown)


install/upgrade/update_to_dev.php
```diff
@@ -178,7 +178,6 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
    $migration->addfield($table, 'plugin_orion_tasks_id', 'integer', ['after' => 'dl_filename']);
    $migration->addKey($table, 'entities_id', 'entities_id');
    $migration->addPostQuery("UPDATE `$table` SET `parse_status` = 'parsed'");
-   $migration->addPostQuery("UPDATE `$table` SET `filename` = CONCAT('" . addslashes(GLPI_DOC_DIR) . "', `filename`)");
```
DIOHz0r: If the update runs a lot of times, GLPI_DOC_DIR wouldn't be concatenated all those times to filename value?

install/installer.class.php
```diff
@@ -509,7 +508,6 @@ protected function upgradeOneStep($toVersion) {
          if (function_exists($updateFunction)) {
             $this->migration->addNewMessageArea("Upgrade to $toVersion");
             $updateFunction($this->migration);
-            $this->migration->executeMigration();
```
btry: workaround a possible bug in GLPI. To be confirmed

btry: I asked to confirm this issue. Or I'll investigate myself in 2 weeks

btry:
@DIOHz0r check it again please.

install/upgrade/update_to_dev.php
```diff
@@ -181,7 +181,16 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
...
+
```
DIOHz0r: I found a solution of the query commented. Remove the folowing add an put again the line 184 with this sentence... 

```sql
UPDATE `$table` SET `filename` = CONCAT('flyvemdm/package/', IF(`filename` REGEXP '^/', SUBSTR(`filename`, 2), `filename`)) WHERE `filename` NOT REGEXP '^flyvemdm/package/';
```
With this you will got all packages with the same strings.

btry: Humm I think you did not saw I completely changed the PR. Check a look on the source code  tab. 

I won't do such kind of SQL requests because this not really the purpose of SQL, and this is harder to maintain. I kept the code, and added some logic to fix the issue. By the  way, as I commented, I found an issue in GLPI which required a workaround for this specific problem.

DIOHz0r: as we talked let's merge the changes an leave the query for those tester that have a problem with the file paths

PR #224, opened 2017-11-25
Fix file upload for glpi ui

btry:
Fix a right check issue when adding a file from the UI

Workaround a var conflict with GLPI (by renaming it). To be reported to GLPI team.


btry:
see https://github.com/glpi-project/glpi/pull/3206

PR #233, opened 2017-12-04
WIP Save name for uploaded package

### LINKED ISSUE #232: Packages uploads name broken
### Expected behaviour
Show the name of the the package on the module

### Actual behaviour
When a package is uploaded the name of it gets lost, at the list there are showed only the ID from the database, also if the package is used in a policy it can't be deleted until if the name value in DB is blank.

### Environment Details
- GLPi version: 9.2
- Flyve MDM version: 2.0.0-dev

1. Upload a package file.
2. Save the details of it.
3. Show the list of packages uploaded.

imagen

DIOHz0r:
## Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

Fixes #232 

## What is the new behavior?
Package name not empty in database and UI.

- [ ] Yes
- [X] No

### DIFF
--- a/inc/package.class.php
+++ b/inc/package.class.php
@@ -131,6 +131,7 @@ class PluginFlyvemdmPackage extends CommonDBTM {
             (isset($options['withtemplate']) && $options['withtemplate'] == 2),
             $this->getType(), -1);
       $fields['filesize']  = PluginFlyvemdmCommon::convertToGiB($fields['filesize']);
+      $fields['name'] = self::getSpecificValueToDisplay('name', $fields['name']);
       $data = [
             'withTemplate' => (isset($options['withtemplate']) && $options['withtemplate'] ? '*' : ''),
             'canUpdate'    => (!$this->isNewID($ID)) && ($this->canUpdate() > 0) || $this->isNewID($ID),
@@ -697,5 +698,26 @@ class PluginFlyvemdmPackage extends CommonDBTM {
       }
 
       return null;
+   }
+
+   /**
+    * Get a specific value to display
+    *
+    * @param string $field
+    * @param array $values
+    * @param array $options
+    * @return string
+    */
+   static function getSpecificValueToDisplay($field, $values, array $options = []) {
+
+      if (!is_array($values)) {
+         $values = [$field => $values];
+      }
+      switch ($field) {
+         case 'name':
+            return ($values['name']) ? $values['name'] : NOT_AVAILABLE;
+            break;
+      }
+      return parent::getSpecificValueToDisplay($field, $values, $options);
    }
 }
--- a/inc/package.class.php
+++ b/inc/package.class.php
@@ -181,7 +181,8 @@ class PluginFlyvemdmPackage extends CommonDBTM {
          return false;
       }
 
-      if (!$this->isFileUploadValid($preparedFile['filename'])) {
+      $localFilename = $preparedFile['filename'];
+      if (!$this->isFileUploadValid($localFilename)) {
          return false;
       }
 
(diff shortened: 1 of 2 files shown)


inc/package.class.php
```diff
@@ -196,6 +197,7 @@ public function prepareInputForAdd($input) {
          if (rename($uploadedFile, $destination)) {
             $input['filesize'] = fileSize($destination);
             $input['dl_filename'] = basename($uploadedFile);
+            $input['name'] = $localFilename;
```
btry: Setting the name of the package is done while running the automatic action which parses it. Be aware the name will be overwritten one or two minute(s) later. Anyway, this is better to have something instead of a ID on display.

inc/package.class.php
```diff
@@ -231,6 +234,7 @@ public function prepareInputForUpdate($input) {
                $filename = pathinfo($destination, PATHINFO_FILENAME);
                $input['filesize'] = fileSize($destination);
                $input['dl_filename'] = basename($destination);
+               $input['name'] = $localFilename;
```
btry: same as previous comment.

btry:
Hi @DIOHz0r, please check my comments. I need to know the purpose before accepting or requesting changes on the PR.

inc/package.class.php
DIOHz0r: but that name is never updated right now, which action do that?

btry: The name is not set when we upload a package. It is intended to be set later, when the package is parsed. Your idea to put the filename looks good for me; it sets a user friendly and temporary value until the parser finish its job. We just need to ensure the value is overwritten by the parser.

DIOHz0r: Ok, this change is not only user friendly, it also fix a bug with "remove application" policy. When the name is empty you can't delete the policy.

btry: you're right. but putting the filename here may lead to an issue if you try to remove an application.

From the agent point of view, we need to send the application name which is similar to *com.something.app**. If t he agent received a filename it will not be able to identify the app to be removed.

Maybe we need to change a few things in the agent to handle the case where an app is not yet parsed or the parser fails.

@rafaelje What are the different ways you can use o identify an application in Android to remove it ?

DIOHz0r: if the main name (com.example.app.vendor) is necesary for Android, I belive that We must change this logic to be compatible with other device's O.S. 

For that we must leave the auto-generated name and savethe correct device file in another colum, don't you think @btry?

btry: Not necessarly: we cannot deploy an apk on iOS, right ? Then we might format the MQTT message depending on the target OS of the package.

before considering changing this, let's ask to @hectorerb if iOS apps have a reversed domain name, and it can be used to detect if an app is installed.

rafaelje: The only way that I can identify the app on Android is with the complete package name

DIOHz0r: > Not necessarly: we cannot deploy an apk on iOS, right ? Then we might format the MQTT message depending on the target OS of the package.

Ok for iOS but what about Windows Phone or blackberry?

hectorerb:
Hi @btry 
Yes, this is the bundleID, it's a unique ID for each application

btry:
Hi @hectorerb 

Then I think we will need to make  the code to g enerate the MQT message more complex. But fist, we need to implement packages for other systems (same or new object ?)

btry:
#243 is designed to replace this PR

DIOHz0r:
Closed in favor of the new PR

PR #243, opened 2017-12-29
Improve UX by allowing to send package name

### LINKED ISSUE #232: Packages uploads name broken

btry:
Proposal to fix #232. See my latest comment on the issue.

### DIFF
--- a/inc/package.class.php
+++ b/inc/package.class.php
@@ -127,9 +127,15 @@ class PluginFlyvemdmPackage extends CommonDBTM {
 
       $twig = plugin_flyvemdm_getTemplateEngine();
       $fields              = $this->fields;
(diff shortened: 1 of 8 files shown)


btry:
this PR replaces the work done on #232 by using an other solution

PR #301, opened 2018-02-02
Add coverage layout

### LINKED ISSUE #300: Header display in coverage section
#### Observed Results:
The coverage section has a different header than the rest of the sections

  
glpi-plugin coverage

#### Expected behavior:

To keep the same display through the different sections

glpi-plugin news

Naylin15:
### Changes description
Fix the header display

### Checklist

### References

Closes #300

### DIFF
--- a/_layouts/coverage.html
+++ b/_layouts/coverage.html
@@ -1,0 +1,7 @@
+---
+layout: container
+---
+
+<div class="table-responsive">
+    {{ content }}
+</div> 


Naylin15:
Ok, I'm having some issues because I can't detect where is the problem for the glpi v1.0.0 space

I solved the logo size and I found where the font changes

Before 

Before

Now

captura de pantalla 2018-02-02 a la s 12 13 59 p m

Naylin15:
I'm not sure but I think it changes automatically with the releases, since the last one is 1.1.0, that's the one it is shown

I'll change the PR to add the coverage layout only

Naylin15:
@DIOHz0r 
the coverage files are being generated with the layout container, it should be changed to coverage (added in this PR) in order make easier any style change for the display

PR #316, opened 2018-02-08
WIP New devices stats report

### LINKED ISSUE #34: maintain history of online status
This allows to compute availability statistics (ie: 56% online) and graph it (per device, per fleet or globally?)

DIOHz0r:
### Changes description

Add a new graph to show the last contacted date and time for agents

### Checklist

### References
Closes flyve-mdm/glpi-plugin#34

### DIFF
--- a/inc/graph.class.php
+++ b/inc/graph.class.php
@@ -245,4 +245,8 @@ class PluginFlyvemdmGraph extends CommonDBTM
          }
       }
    }
+
+   public function showAgentOnlineStats() {
+
+   }
 }
--- a/inc/graph.class.php
+++ b/inc/graph.class.php
@@ -247,6 +247,120 @@ class PluginFlyvemdmGraph extends CommonDBTM
    }
 
    public function showAgentOnlineStats() {
+      global $DB;
 
+      $domain = $range = $serie = [];
+      $FlyvemdmAgent = PluginFlyvemdmAgent::getTable();
+      $logTable = Log::getTable();
+      $query = "SELECT l.`new_value` AS online_date, f.`name` AS agent_name, f.`id` AS agent_id 
+        FROM " . $logTable . " AS l 
+        INNER JOIN " . $FlyvemdmAgent . " AS f ON l.`items_id` = f.`id`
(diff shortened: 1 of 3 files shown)


inc/graph.class.php
```diff
@@ -277,4 +277,147 @@ private function checkEmptyLabels(&$labels) {
...
       }
    }
+
+   public function showAgentOnlineStats() {
+      global $DB;
+
+      $domain = $range = $serie = [];
+      $FlyvemdmAgent = PluginFlyvemdmAgent::getTable();
+      $logTable = Log::getTable();
+      $query = "SELECT l.`new_value` AS online_date, f.`name` AS agent_name, f.`id` AS agent_id
```
btry: If you introduce queries un the plugin, use the query builder. RAW SQL queries is deprecated.

http://glpi-developer-documentation.readthedocs.io/en/master/devapi/database/index.html

btry (changes requested):
Please change the quiery as I recommended.

In the mean  time I'll check how it looks (and add a screenshot if nice to show)

inc/graph.class.php
DIOHz0r: done it on new commit

btry:
@DIOHz0r 

if we have hundreds  or thousands of agents, I think the graph will  be unreadable. I think it would  be more interesting to count the number of devices online per hour of a day for example. But again, this would be useful only for the current day, or the few last ones.

Moreover the current query tracks the is_online changes, not their status at some point of the timeline.

As is, I'm not convinced the admin will get useful information from this graph. I think it would be more interesting to get the online ratio per hour for the last few days (week for example ?)

Also the logs table may be deleted by a GLPI automatic action every X days (30 by default) we have a limited ability to look in the past.

btry:
Also you should filter by search option 11, not 8. 11 is the flag is_online, and it is t imestamped in the logs.

DIOHz0r:
I'm conscious about having a lot of data on the report. The query is incomplete and the idea is show only a range of days maybe 5 days by default. Let's try first to show the data and then create the default range of info to show

btry:
What about the other points I talked about. Do you plan to show every agent ? Do you plan to show dots for is_online changes ?

ajsb85 (dismissed):

PR #324, opened 2018-02-14
Optimize speed by running CS check only once

btry:


btry:
Dear, bot, I agree, but really, read the title :) it is self explanatory

PR #327, opened 2018-02-14
Patch for before script code

DIOHz0r:

### Changes description

Update code for Travis before_script to not remove the composer.lock and been able to test for different PHP versions env according to this post https://hannesvdvreken.com/2015/01/18/composer-ignore-platform-reqs-flag/

### Checklist

### References

Patch related to PR #325

### DIFF
--- a/tests/before_script.sh
+++ b/tests/before_script.sh
@@ -4,6 +4,13 @@
 # Before script for Travis CI
 #
 
+# config composer
+if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then
+  mkdir ~/.composer -p
+  touch ~/.composer/composer.json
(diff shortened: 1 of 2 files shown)


tests/before_script.sh
```diff
@@ -29,5 +34,5 @@ if [[ $GLPI_BRANCH == "9.2.1" ]] ; then patch -p1 --batch < plugins/flyvemdm/tes
 
 # prepare plugin to test
 cd plugins/flyvemdm 
-composer install --no-interaction
+composer install --no-interaction --ignore-platform-reqs
```
btry: I think we should not add this argument, as we discussed. Please,  try to revert this line and let's see how the tests run on  this PR.

DIOHz0r: ok

PR #317, opened 2018-02-09
WIP Add comment field in File web page

Oyabi:
### Changes description

Add a comment field in File webpage and File list view.
Useful in the case where several files have the same name in order to differentiate them.

### Checklist

Please check if your PR fulfills the following specifications:

- [ ] Tests for the changes have been added
- [ ] Docs have been added/updated

### References

Closes #N/A

PR #328, opened 2018-02-14
Add comment field

btry:
Rebase #317 and enhancements

WIP : need to check unit tests for prepareInputForAdd and prepareInputForUpdate of the PluginFlyvemdmFile

### Changes description

Add a comments field for files

### Checklist

### DIFF
--- a/front/file.form.php
+++ b/front/file.form.php
@@ -48,6 +48,10 @@ if (isset($_POST['add'])) {
    $flyveFile->check(-1, CREATE, $_POST);
    $flyveFile->add($_POST);
    Html::back();
+} else if (isset($_POST['update'])) {
+   $flyveFile->check(-1, UPDATE, $_POST);
+   $flyveFile->update($_POST);
+   Html::back();
 } else {
    $flyveFile->check($_GET['id'], READ);
    Html::header(
--- a/inc/file.class.php
+++ b/inc/file.class.php
@@ -75,38 +75,9 @@ class PluginFlyvemdmFile extends CommonDBTM {
    }
 
    public function prepareInputForAdd($input) {
-      // Find the added file
-      if (!isAPI()) {
-         // from GLPI UI
-         if (isset($_POST['_file'][0]) && is_string($_POST['_file'][0])) {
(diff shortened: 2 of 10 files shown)


front/file.form.php
```diff
@@ -48,6 +48,10 @@
    $flyveFile->check(-1, CREATE, $_POST);
    $flyveFile->add($_POST);
    Html::back();
+} else if (isset($_POST['update'])) {
+   $flyveFile->check(-1, UPDATE, $_POST);
+   $flyveFile->update($_POST);
+   Html::back();
```
btry: updating a file was not handled

inc/file.class.php
```diff
@@ -212,14 +214,13 @@ public function prepareInputForUpdate($input) {
          if ($filename != $this->fields['source']) {
             unlink(FLYVEMDM_FILE_PATH . "/" . $this->fields['source']);
          }
+         // File updated, then increment its version
+         $input['version'] = $this->fields['version'] + 1;
```
btry: nice bug where the version was not udated . . post increment instead of pre increment 👎

inc/file.class.php
```diff
@@ -68,70 +68,31 @@ public static function getMenuPicture() {
...
-         Session::addMessageAfterRedirect(__('File uploaded without name', "flyvemdm"));
+         Session::addMessageAfterRedirect(__('File uploaded without name', 'flyvemdm'));
          return false;
       }
       $input['source'] = $input['entities_id'] . "/" . uniqid() . "_" . basename($uploadedFile);
       $destination = FLYVEMDM_FILE_PATH . "/" . $input['source'];
       $this->createEntityDirectory(dirname($destination));
       if (!rename($uploadedFile, $destination)) {
-         if (!is_writable(dirname($destination))) {
+         if (!file_exists(dirname($destination))) {
```
DIOHz0r: I don't like to use file_exists function for check if a file is uploaded because that function returns true even if the file doesn't exist. It's behavior is to check against files and dirs and if the source input is empty it will return true, so it's better to use is_file or is_readable IMO

PR #337, opened 2018-02-20
Maximize wizard use of the wizard

### LINKED ISSUE #333: Plugin prerequisites duplicates wizard
When the files of the plugin are installed, we cannot install and enable it without eanbling the API and the email notifications.

This duplicates the wizard pages to configure them, and is less user friendly. 

We need to remobe these checks and show a warning in  the Flyve MDM menu page. The same for  the dashboard (issue to create)

btry:
### Changes description

Add a missing step in the wizard, move the GLPI configuration requirement checks

If GLPI seems not properly configured, a text is displayed above the UI. May be enhanced for a nicer display.

image

### References

Closes #333

### DIFF
--- a/tpl/menu.html
+++ b/tpl/menu.html
@@ -1,3 +1,8 @@
+{% if configuration == false %}
+<center>
+	<p>{{ __('GLPI is not properly configured. Please check the general settings, the API settings and the notification settings.', 'flyvemdm' )}}</p>
+</center>
+{% endif %}
 <center>
    <ul id="plugin_flyvemdm_menu">
       {% for menuName, menuItem in menu %}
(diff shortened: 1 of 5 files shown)


tpl/menu.html
```diff
@@ -1,3 +1,8 @@
+{% if configuration == false %}
+<center>
+	<p>{{ __('GLPI is not properly configured. Please check the geleral settings, the API settings and the notification settings.', 'flyvemdm' )}}</p>
```
Naylin15: Hi, @btry 

There is a typo, "geleral"

btry:
Thanks ! last commit amended and pushed

PR #353, opened 2018-02-23
Add filesize information

DIOHz0r:
### Changes description

Upgrade code to show filesize info for files and packages

### Checklist

Please check if your PR fulfills the following specifications:

- [x] Tests for the changes have been added
- [x] Docs have been added/updated

### References

Closes #176

PR #361, opened 2018-03-07
Unit tests updates

DIOHz0r:
### Changes description

This changes adds new unit test and fix issues related with the size info for packages and files

### Checklist

### References

Issue #353


tests/suite-unit/PluginFlyvemdmPackage.php
```diff
@@ -37,13 +37,84 @@ class PluginFlyvemdmPackage extends CommonTestCase {
...
+    * @param $method
+    */
+   public function afterTestMethod($method) {
+      switch ($method) {
+         case 'testShowForm':
+         case 'testPrepareInputForAdd':
+         case 'testAddNeededInfoToInput':
+         case 'testPrepareInputForUpdate':
+            parent::afterTestMethod($method);
+            \Session::destroy();
```
btry: Is this call really useful ? Destriying sessions is handled when we need to build a new one in the test tools package

tests/suite-unit/PluginFlyvemdmPackage.php
```diff
@@ -37,13 +37,84 @@ class PluginFlyvemdmPackage extends CommonTestCase {
...
+      }
+   }
+
+   /**
+    * @return object
+    */
+   public function createInstance() {
+      $instance = $this->newTestedInstance();
+      return $instance;
+   }
```
btry: remove this method. it is useless to make a method for a single line of code.

lib/GlpiLocalesExtension.php
```diff
@@ -61,4 +61,21 @@ public function getFunctions() {
...
+
+   /**
+    * Format a size passing a size in octet
+    * @param int $number
+    * @return string
+    */
+   public function fileSizeFilter($number)
+   {
+      return Toolbox::getSize($number);
+   }
```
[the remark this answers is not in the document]
DIOHz0r: I do not because if we need to manipulate the value (which happens on some hooks) is more easy if there are the raw value also the data in bytes are sent to the web dashboard and they handle it in diffent way. 

And as it says on the Yii best practices for the views... "views should mainly contain presentational code, such as HTML, and simple PHP code to traverse, format and render data;" ;)

tests/suite-unit/PluginFlyvemdmPackage.php
DIOHz0r: Is useful for ensure that we are cleaning up the session to start a new one on the next test with the same values. The CommonTestCase hasn't have a session destroy on the login action to ensure that we would have always the same session vars.

tests/suite-unit/PluginFlyvemdmPackage.php
DIOHz0r: This is used on testShowForm and testAddNeededInfoToInput, to have the same instance for testing on each test

btry: Understood

tests/suite-unit/PluginFlyvemdmPackage.php
btry: OK, understood

tests/suite-unit/PluginFlyvemdmPackage.php
btry: tests must be independent. Each test must use its own instance.

Moreover, atoum uses child processes : one child is created per test method, then it simply dies then you actually have a distinct instance for each test method.

lib/GlpiLocalesExtension.php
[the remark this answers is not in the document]
btry: ok, let's try. Next step is to find how to drop  the |raw in many variables of the templates .. HTML code  generated for dropdowns, ...

PR #364, opened 2018-03-12
Send remove message to device

### LINKED ISSUE #294: Create a Policy/removeApp topic with message empty
#### Observed Results:

- Create a new fleet
- Create a new Files
- Add policy "Deploy File"
- Remove the policy "Deploy File"
- The plugin automatically adds a new policy "Remove file" but the value of the message send to the agent is empty

captura de pantalla 2018-01-31 a la s 3 52 10 p m

#### Expected behavior:

- The message of the "Remove file" policy will be the document to remove

DIOHz0r:
### Changes description

Send correct remove text applications to agents.

### Checklist

### References

Closes #294 

Depends on #369


tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,130 @@
...
+    * @param $data
+    * @param $expected
+    */
+   public function testIntegrityCheck($data, $expected) {
+      list($policy) = $this->createNewPolicyInstance();
+      $success = $policy->integrityCheck($data[0], $data[1], $data[2]);
+      $this->boolean($success)->isEqualTo($expected[0]);
+      if (!$expected[0] && isset($expected[1])) {
+         $this->string($_SESSION["MESSAGE_AFTER_REDIRECT"][0][0])->isEqualTo($expected[1]);
+         unset($_SESSION["MESSAGE_AFTER_REDIRECT"]); // to clear the buffer
```
btry: As you often use beforeTestMethod() to setup things, I think you should use afterTestMethod() to unset this value.

btry: then OK

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,233 @@
...
+         $policyData->getFromDBBySymbol('removeApp');
+      }
+      list($policy) = $this->createNewPolicyInstance($policyData);
+      $mockInstance = $this->newMockInstance('\PluginFlyvemdmFleet');
+      $mockInstance->getMockController()->getID = 1;
+      $result = $policy->apply($mockInstance, $data[0], $data[1], $data[2]);
+      $this->boolean($result)->isEqualTo($expected['return']);
+      if (!$expected['return'] && isset($expected['message'])) {
+         $this->string($_SESSION["MESSAGE_AFTER_REDIRECT"][1][0])->isEqualTo($expected['message']);
+         unset($_SESSION["MESSAGE_AFTER_REDIRECT"]); // to clear the buffer
```
btry: Coming again here.

the unset() is done only when the test method is successful. If it fails, it may break an other execution of the method for the next data set from the data provider.

in the if(), copy ```$_SESSION["MESSAGE_AFTER_REDIRECT"][1][0]``` in a variable, next unset it, next test the variable.

inc/policydeployapplication.class.php
```diff
@@ -118,7 +118,8 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
       $fleetId = $fleet->getID();
       $task = new PluginFlyvemdmTask();
       $rows = $task->find("`plugin_flyvemdm_fleets_id` = '$fleetId'
-            AND `itemtype` = '$itemtype' AND `items_id` = '$itemId'", '', '1');
+            AND `plugin_flyvemdm_policies_id` = '" . $this->policyData->getID() . "' 
+            AND `items_id` = '$itemId'", '', '1');
```
btry: What was the problem by checking the itemtype instead if the policy id ?

inc/policyremoveapplication.class.php
```diff
@@ -56,14 +56,10 @@ public function __construct(PluginFlyvemdmPolicy $policy) {
     * @return bool
     */
    public function integrityCheck($value, $itemtype, $itemId) {
-      if ($value == '') {
+      if (!$itemId) {
          Session::addMessageAfterRedirect(__('An application ID is required', 'flyvemdm'));
```
btry: The application ID of the message is maybe misleading. This is not an ID from the database but a string like "com.vendor.application" . The value is mandatory because the a gent needs  this to determine the app to uninstall.

inc/policyremoveapplication.class.php
```diff
@@ -90,13 +87,91 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
...
    public function getMqttMessage($value, $itemtype, $itemId) {
-      if (! $this->integrityCheck($value, $itemtype, $itemId)) {
+      if (!$this->integrityCheck($value, $itemtype, $itemId)) {
          return false;
       }
 
+      // force the itemtype for packages
+      if (!($itemtype instanceof PluginFlyvemdmPackage)) {
+         $itemtype = new PluginFlyvemdmPackage();
+      }
```
btry: $itemtype is a string, not an instance

Also, if the caller sends an unexpected itemtype, it is better to give up with a error code or exception rather than "guessing" what was wanted.

inc/policyremoveapplication.class.php
```diff
@@ -90,13 +87,91 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
...
+
+      $itemtype->getFromDB($itemId);
```
btry: if you want to get the item for any reason, prefer $item = new Itemtype()

inc/policyremoveapplication.class.php
```diff
@@ -90,13 +87,91 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
...
+    * @param int $itemId
+    * @return bool
+    */
+   public function apply(PluginFlyvemdmFleet $fleet, $value, $itemtype, $itemId) {
+      $this->canApply = false;
+
+      // force the itemtype for packages
+      if (!($itemtype instanceof PluginFlyvemdmPackage)) {
+         $itemtype = new PluginFlyvemdmPackage();
+      }
```
btry: same as above. itemtype is a string (a type), not an instance

inc/policydeployapplication.class.php
DIOHz0r: The policies of  deploy and removal app now saves the class name string so to identify them I'm searching for the policy id.

inc/policyremoveapplication.class.php
DIOHz0r: Now the Id is saved on the task and is mandatory.

inc/policyremoveapplication.class.php
DIOHz0r: That's why I force it, I don't care the value of the itemtype this policies works only for Packages

inc/policydeployapplication.class.php
```diff
@@ -191,16 +192,11 @@ public function unapply(PluginFlyvemdmFleet $fleet, $value, $itemtype, $itemId)
...
-         'plugin_flyvemdm_policies_id' => $policyData->getID(),
-         'value'                       => $packageName,
-         '_silent'                     => true,
-      ])) {
-         return false;
+      $package = new PluginFlyvemdmPackage();
+      if ($package->getFromDB($itemId)) {
+         $policyFactory = new PluginFlyvemdmPolicyFactory();
+         $removeApp = $policyFactory->createFromPolicy($policyData);
+         $removeApp->apply($fleet, $package->getField('package_name'), $package, $itemId);
```
btry: 3rd argument must be PluginFlyvemdmPackage::class or equivalent. A string is expected, not an object.

(by the way we need to fix the docblock describing this methid in PluginFlyvemdmPolicyBase::apply() )

inc/policydeployapplication.class.php
btry: Ok for this; I think this also fixes a bug.

inc/policyremoveapplication.class.php
btry: as we discussed today, this if() block becomes useless, there is no need to save an itemtype or a items_id on the DB for app removal.

inc/policyremoveapplication.class.php
```diff
@@ -90,13 +91,58 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
...
+
+      // the itemId wasn't send and the itemtype is new
+      if ($itemtype->isNewItem() && !$itemId) {
+         Session::addMessageAfterRedirect(__('An application ID is required', 'flyvemdm'),
+            false, ERROR);
+         return false;
+      }
```
btry: this is useless  too

inc/policyremoveapplication.class.php
```diff
@@ -90,13 +91,58 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
...
+
+      // the itemId is send but the itemtype is new, lest's try to load the info of the package
+      if ($itemtype->isNewItem() && $itemId) {
+         if (!$itemtype->getFromDB($itemId)) {
+            Session::addMessageAfterRedirect(__('The application does not exists', 'flyvemdm'),
+               false, ERROR);
+            return false;
+         }
+      }
```
btry: this block is useless too

inc/policyremoveapplication.class.php
```diff
@@ -90,13 +91,58 @@ public function unicityCheck($value, $itemtype, $itemId, PluginFlyvemdmFleet $fl
...
+
+      // the package is loaded, let's create the task
+      $task = new PluginFlyvemdmTask();
+      if (!$task->add([
+         'plugin_flyvemdm_fleets_id'   => $fleet->getID(),
+         'plugin_flyvemdm_policies_id' => $this->getPolicyData()->getID(),
+         'value'                       => $itemtype->getField('package_name'),
```
btry: use  the value received as argument here.

Also, for readability, I think it is better to call $task->add()  out of the if () condition. You can write 
```php
if ($task->isNewItem()) {
   return false;
}

File: inc/task.class.php [hunk 12]
```diff
@@ -197,13 +197,6 @@ public function prepareInputForAdd($input) {
 
       // Check the policy may be applied to the fleet and the value matches requirements
       if (!$this->policy->canApply($this->fleet, $input['value'], $input['itemtype'],
-         $input['items_id'])) {
-         Session::addMessageAfterRedirect(__('The requirements for this policy are not met',
```
btry: I think this message is more appropriate than the other L208/201

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,279 @@
...
+      ];
+   }
+
+   /**
+    * @dataProvider integrityCheckProvider
+    * @tags testIntegrityCheck
+    * @param array $data
+    * @param array $expected
+    */
+   public function testIntegrityCheck(array $data, array $expected) {
```
btry: Please split in the provider and here the argument $data into 3 variables with a meaningful name

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,279 @@
...
+      ];
+   }
+
+   /**
+    * @dataProvider unicityCheckProvider
+    * @tags testUnicityCheck
+    * @param array $data
+    * @param array $expected
+    */
+   public function testUnicityCheck(array $data, array $expected) {
```
btry: same here: more meaningful variable names

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,279 @@
...
+      ];
+   }
+
+   /**
+    * @dataProvider applyProvider
+    * @tags testApply
+    * @param array $data
+    * @param array $expected
+    */
+   public function testApply(array $data, array $expected) {
```
btry: same here

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,279 @@
...
+      $mockInstance->getMockController()->getField = -1;
+      $result = $policy->showValue($mockInstance);
+      $this->string($result)->isEqualTo(NOT_AVAILABLE);
+   }
+
+   /**
+    * Create an application (directly in DB) because we are not uploading any file
+    * @return \PluginFlyvemdmPackage
+    */
+   private function createAppInDB() {
```
btry: This method should move in tests/src/Flyvemdm/TestingCommonTools.php . After we merge this PR, let's create a PR to share this method with other tests needing it.

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,279 @@
...
+
+      $file = new \PluginFlyvemdmPackage();
+      $file->getFromDB($DB->insert_id());
+      $this->boolean($file->isNewItem())->isFalse();
+
+      return $file;
+   }
+
+   /**
+    * Create an application (directly in DB) because we are not uploading any file
```
btry: copy pasted description

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
```diff
@@ -0,0 +1,279 @@
...
+    * @return \PluginFlyvemdmPackage
+    */
+   private function createFleetInDB() {
+      global $DB;
```
btry: I think there is no need to create a fleet with a direct SQL query. I think other unit tests create fleets the regular way. If I'm right, let's mutualize this method in TestingCommonTools, and in a code refactor PR (see previous comment), let's mutualize it.

tests/suite-unit/PluginFlyvemdmPolicyRemoveapplication.php
btry: give subkeys to the key **data**

example:
```php
'data'     => ['package_name' => 'fake.lorem.ipsum.package', 'itemtype' => '', items_id => 0],
```

and the unit tests will be more readable

btry:
@DIOHz0r The PR is ok for me. Please squash all commits to have a clean history in t he develop branch, and I'll approve it. I let you the honor to click the merge button :)

PR #367, opened 2018-03-15
Fix deprecated GLPI methods calls

### LINKED ISSUE #365: getFromDBByQuery deprecated in GLPI 9.3
at least one PR fails with GLPI 9.3 (currently branch master) because getFromDBByQuery is deprecated.

We need to find all calls to this method and replace them by $DB->request(), in order to make tests work again. 

This is high priority because other tests on other PRs depend on this (and not so hard)

btry:
fix #365 

This PR requires https://github.com/glpi-project/glpi/pull/3724

### DIFF
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -556,6 +556,8 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
     * @return bool : true if item need to be deleted else false
     */
    public function pre_deleteItem() {
+      global $DB;
+
       $success = false;
 
       // get serial of the computer
@@ -610,7 +612,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
       // Delete the MQTT user for the agent
       if (!empty($serial)) {
          $mqttUser = new PluginFlyvemdmMqttuser();
-         if ($mqttUser->getFromDBByQuery("WHERE `user` = '$serial'")) {
+         if ($mqttUser->getFromDBByCrit(['user' => $serial])) {
             if (!$mqttUser->delete(['id' => $mqttUser->getID()], true)) {
                Session::addMessageAfterRedirect(__('Failed to delete MQTT user for the device', 'flyvemdm'));
                return false;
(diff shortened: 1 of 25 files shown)


inc/agent.class.php
```diff
@@ -928,11 +930,23 @@ public function getByTopic($topic) {
             if (strlen($serial)) {
                $computerTable = Computer::getTable();
                $agentTable = self::getTable();
-               return $this->getFromDBByQuery("LEFT JOIN `$computerTable` `c` ON (
-                     `c`.`id` = `$agentTable`.`computers_id`
-                     )
-                     WHERE `$agentTable`.`entities_id`='$entity' AND `c`.`serial` = '$serial'"
-               );
+               $result = $DB->request([
```
DIOHz0r: You forgot to add the WHERE arguments on the new request

inc/config.class.php
```diff
@@ -188,7 +188,7 @@ public function addDocumentTypes() {
 
       foreach ($extensions as $extension => $name) {
          $documentType = new DocumentType();
-         if (!$documentType->getFromDBByQuery("WHERE LOWER(`ext`)='$extension'")) {
+         if (!$documentType->getFromDBByCrit(['ext' => $extension])) {
```
DIOHz0r: Doesn't the lower matters?

btry: well... no. Mysql is case insensitive when comparing strings !

Try to select an entry of glpi_documenttypes with a where on the column ext, and with an existing entry you type un upper case...

inc/agent.class.php
btry: +1

inc/agent.class.php
```diff
@@ -918,26 +920,34 @@ public function getTopic() {
...
+               ]
+            ]
+         ],
+         'WHERE' => [
+            'AND' => [
+               PluginFlyvemdmAgent::getTable() . '.' . Entity::getForeignKeyField() => $entity,
+               Computer::getTable() . '.serial' => $serial
+            ]
+         ]
+      ]);
```
btry: refactored to make code easier to read by reversing the if()

PR #385, opened 2018-03-23
Fix manifest malfunction

btry:
Flyve manifest topic had 2 problems
* infinite loop when updating the version if Flyve in the manifest
* leading slash in the topic not removed here in  the previous leading slash removal PR


inc/mqtthandler.class.php
```diff
@@ -121,23 +121,19 @@ public function publish(\sskaje\mqtt\MQTT $mqtt, \sskaje\mqtt\Message\PUBLISH $p
       if (isset($mqttPath[3])) {
          if ($mqttPath[3] == "Status/Ping") {
```
ajsb85: Why not switch statement?

PR #386, opened 2018-03-23
WIP Introduce dependency injection

### LINKED ISSUE #171: Use a dependency container
GLPI and its plugins son't use a dependency container. THis impacts code quality, and unit tests.

I think it is possible to workaround this in the scope of the plugin by using a dependency container stored in a global variable (we cannot do better than a global for now). 

This global will provide services to classes of the plugin and help to greatly break tight couplig between classes, and unlock better unit  tests.

IMO this is an urgent task to begin in december 2017 / january 2018. I think this will help to speedup development too.

I searched and selected this container

https://github.com/silexphp/Pimple

Note: no milestone yet.

btry:
Really needed to improve unit tests coverage

This PR tries to follow the future practices of GLPI in WIP: https://github.com/glpi-project/glpi/pull/3450

Will fix flyve-mdm/glpi-plugin#171 

### Estimated time

(since July 20th)

|Assignee|:tomato:|
|:---|:---:|
|@diohz0r |12|

### DIFF
--- a/tests/src/Flyvemdm/Tests/CommonTestCase.php
+++ b/tests/src/Flyvemdm/Tests/CommonTestCase.php
@@ -587,4 +587,17 @@ class CommonTestCase extends GlpiCommonTestCase {
          'Policy/defaultStreamType',
       ];
    }
+
+   /**
+    * Prepares the dependency container
+    * @param array $definition the definitions of the DI
+    */
+   public function setupDiContainer($definition) {
+      global $pluginFlyvemdmContainer;
+
+      $containerBuilder = new \DI\ContainerBuilder();
+      $containerBuilder->useAnnotations(true);
+      $containerBuilder->addDefinitions($definition);
+      $pluginFlyvemdmContainer = $containerBuilder->build();
+   }
 }
--- a/ajax/geolocation.php
+++ b/ajax/geolocation.php
@@ -55,7 +55,7 @@ if (isset($_REQUEST['endDate']) && !empty(trim($_REQUEST['endDate']))) {
    $endDate = $endDate->format('Y-m-d H:i:s');
 }
 
-$geolocation = new PluginFlyvemdmGeolocation();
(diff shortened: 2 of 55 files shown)


tests/src/Flyvemdm/Tests/CommonTestCase.php
```diff
@@ -144,4 +144,17 @@ protected function agentFromInvitation(
...
+
+   /**
+    * Prepares the dependency container
+    * @param array $definition the definitions of the DI
+    */
+   public function setupDiContainer($definition) {
+      global $pluginFlyvemdmContainer;
+
+      $containerBuilder = new \DI\ContainerBuilder();
+      $containerBuilder->useAnnotations(true);
```
DIOHz0r: I'm not in favor of use annotations and I didn't see that you aren't using it. 

We must start thinking as services for future versions and PHP-DI recommends do not use annotation for injecting dependencies on services ([ref: best practices.](http://php-di.org/doc/best-practices.html#writing-services))

btry: It is enabled, but not used. I followed the configuration of PHP-DI in the PR for GLPI.

btry: I agree about services, but this will not help us enough to write better tests. I want to use DI to access the $DB object, but we cannot solve this in the plugin only. We need DI in GLPI (because of code in CommonDBTM)

DIOHz0r (dismissed):

btry:
Unit tests are OK now. 

Next step is to replace ->make() calls by ->get() to follow PSR-11 (interface in this spec does not knows make() )

As we absolutely need to have fresh instances of objects in some cases, I think we may use factories to use the method get() and have fresh objects. 

http://php-di.org/doc/php-definitions.html#factories

DIOHz0r:
Hi, this PR has been updated based on v2.0.0-rc.1

ajsb85 (approved):

btry:
obsoleted. Nex attempt of  DI system in progress in GLPI

PR #389, opened 2018-03-27
Use migration object to update configuration

btry:
use GLPI API when possible !

### DIFF
--- a/install/upgrade/update_to_dev.php
+++ b/install/upgrade/update_to_dev.php
@@ -53,22 +53,17 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
    ]);
    $profileRight->updateProfileRights($profiles_id, $newRights);
 
-   Config::setConfigurationValues('flyvemdm', [
+   $migration->setContext('flyvemdm');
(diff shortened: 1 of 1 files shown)


install/upgrade/update_to_dev.php
```diff
@@ -53,22 +53,17 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
...
-   Config::setConfigurationValues('flyvemdm', [
+   $migration->setContext('flyvemdm');
+   $migration->addConfig([
       'default_agent_url'     => PLUGIN_FLYVEMDM_AGENT_DOWNLOAD_URL,
       'show_wizard'           => '0',
       'debug_save_inventory'  => '0',
+      'android_bugcollecctor_url' => '',
+      'android_bugcollector_login' => '',
+      'android_bugcollector_passwd' => '',
+      'invitation_deeplink' => PLUGIN_FLYVEMDM_DEEPLINK,
```
btry: addConfig() creates a configuration value if it does not exists yet in glpi_configs table.

PR #396, opened 2018-03-30
fix fleet deletion & drop dead code

btry:
This PR is not urgent

### DIFF
--- a/inc/fleet.class.php
+++ b/inc/fleet.class.php
@@ -162,16 +162,6 @@ class PluginFlyvemdmFleet extends CommonDBTM implements PluginFlyvemdmNotifiable
     * @return bool : true if item need to be deleted else false
     */
    public function pre_deleteItem() {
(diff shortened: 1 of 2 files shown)


inc/fleet.class.php
```diff
@@ -162,16 +162,6 @@ public function prepareInputForUpdate($input) {
     * @return bool : true if item need to be deleted else false
     */
    public function pre_deleteItem() {
-      // check if fleet being deleted is the default one
```
DIOHz0r: If your are allowing to remove the "not managed fleet", then the message on https://github.com/btry/glpi-plugin/blob/c829ad7c2c4e32b6f4292e1ebd9c4ab8a93883c7/inc/fleet.class.php#L185 should be changed to "Could not move all devices to the default fleet"

btry: No: the code I'm removing actually does nothing.

PR #401, opened 2018-04-03
Allow recursion change for non default fleets

btry:
One may change the recursion of a managed fleet


btry:
@DIOHz0r I added unit tests. The PR is OK now

Issue #344, opened 2018-02-21
Display problem in entity configuration 


btry:
The inheritance of entity settings is not handled correctly in the UI
  
#### Observed Results:

![image](https://user-images.githubusercontent.com/14139801/36491660-9836c918-172b-11e8-9307-69cce80429fd.png)

#### Expected behavior:

![image](https://user-images.githubusercontent.com/14139801/36491703-b3d46ba8-172b-11e8-88da-ac5a6b698570.png)

DIOHz0r:
Hi, could you indicate the steps or the environment to reproduce this error? I've tried on my local develop environment without that '-2' values

btry:
- Create an entity child of the root entity.
- In the root entity, setup any text to fill one or several fields
- edit  the sub entity and view the support informations. You should see the value -2 which has a special meaning. I think we need improvements here to "catch" this value and display the value of the parent entity.

DIOHz0r:
I did your steps and I don't get those values, I'm getting the root entities values 
![imagen](https://user-images.githubusercontent.com/1426313/36787687-54c67878-1c61-11e8-8bba-a88a2ab60d03.png)

btry:
Indeed, I think the entity I viewed was created when the plugin was disabled or something like that. I found this while working on an other plugin.

Let's close it.

PR #427, opened 2018-04-12
Debug setting

### LINKED ISSUE #424: Add debug setting
#### Observed Results:
Add debug setting to select between the URLs of the beta and production version of the Agent

  
#### Expected behavior:
New debug setting

Naylin15:
### Changes description
Add a new tab for the debug setting
More changes to come

### Checklist

### References

Closes #424


inc/entityconfig.class.php
```diff
@@ -370,6 +369,11 @@ public function showFormForEntity(Entity $item) {
       $canedit = static::canUpdate();
 
       $fields = $this->fields;
+      $fields['download_url'] = Dropdown::showFromArray('download_url',
+            [PLUGIN_FLYVEMDM_AGENT_DOWNLOAD_URL => __('MDM Agent Play Store'),
+            PLUGIN_FLYVEMDM_AGENT_BETA_DOWNLOAD_URL => __('MDM Agent  Beta Testing'),],
```
btry: 2 spaces before **Beta testing** !

Naylin15:
The cli install --dev part of the issue #424 is ready, it now changes the download url to the beta testing in root entity only

btry:
@Naylin15 I noticed a few changes in the twig templates I did not expected. I think you encountered a bug while developing and testing the feature. Maybe this one: https://github.com/flyve-mdm/glpi-plugin/issues/344

I'll add to you branch a unit test.

tpl/entity_entityconfig.html.twig
```diff
@@ -3,55 +3,78 @@
       {{ __('Support name', 'flyvemdm')  }}
    </td>
    <td>
-      <input name="support_name" value="{{ entityConfig.support_name }}" />
-      {% if entityConfig._support_name == -2 %}
+    {% if entityConfig._support_name == -2 %}
+      <input name="support_name" />
```
btry: I'm reviewing in deep to understand why you changed this (https://github.com/flyve-mdm/glpi-plugin/pull/427#issuecomment-382663086) and I found this change is not necessary.

If the value is inherited, then ```entityConfig.support_name``` is an empty string. Then no need to move the <input> tag in the % IF %. Or did I missed something ?

tools/cli_install.php
```diff
@@ -156,6 +158,17 @@
...
 
+if($dev) {
+   $entityConfig = new PluginFlyvemdmEntityConfig();
+   $entityConfig->getFromDBByCrit([
+      'entities_id' => '0',
+   ]);
+   $devquery = "UPDATE `glpi_plugin_flyvemdm_entityconfigs`
+               SET `download_url` = '" . PLUGIN_FLYVEMDM_AGENT_BETA_DOWNLOAD_URL . "'
+               WHERE `glpi_plugin_flyvemdm_entityconfigs`.`entities_id` = 0";
+   $DB->query($devquery);
```
btry: Why not update the entityConfig with the following code ? 
```php
   $entityConfig->update([
      'id'           => $entityConfig->getID(),
      'download_url' => PLUGIN_FLYVEMDM_AGENT_BETA_DOWNLOAD_URL
   ]);
```

(If you don't see any problem with that, I'm ready to commit to your pr, then merge.)

tools/cli_install.php
```diff
@@ -163,6 +163,10 @@
    $entityConfig->getFromDBByCrit([
       'entities_id' => '0',
    ]);
+   $entityConfig->update([
+      'id'           => $entityConfig->getID(),
+      'download_url' => PLUGIN_FLYVEMDM_AGENT_BETA_DOWNLOAD_URL
+   ]);
    $devquery = "UPDATE `glpi_plugin_flyvemdm_entityconfigs`
```
Naylin15: Hi @btry 

Shouldn't these lines ($devquery) be removed from the code? 

Since the new code makes the update

tools/cli_install.php
btry: You're right. This was a draft I forgot to drop when I commented this part of code... Removing

tools/cli_install.php
btry: You're right. This was a draft when I commented this part of code, and I forgot to drop ip... I'll clean it up

PR #428, opened 2018-04-13
Do not silently fail to update entity config

### LINKED ISSUE #426: PluginFlyvemdmEntityconfig update
#### Observed Results:

I am sending to update the property 'agent_token_life' of the itemtype PluginFlyvemdmEntityconfig with the id. But I find the following problem:

The backend is returning a 200, which I interpret as a correct answer. But the reality is that this value is not being updated.

captura de pantalla 2018-04-12 a la s 1 14 03 p m
captura de pantalla 2018-04-12 a la s 1 14 09 p m
captura de pantalla 2018-04-12 a la s 1 13 55 p m

  
#### Expected behavior:

Update the value if possible, otherwise return an error message

btry:
fix #426

### Checklist

Closes #426

### DIFF
--- a/inc/entityconfig.class.php
+++ b/inc/entityconfig.class.php
@@ -127,19 +127,32 @@ class PluginFlyvemdmEntityConfig extends CommonDBTM {
     * @return array|false
     */
    public function prepareInputForUpdate($input) {
+      $failure = false;
+
       if (!Session::haveRight(static::$rightname,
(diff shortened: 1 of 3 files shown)


btry:
@DIOHz0r Noticed I added in this PR a helper method to run tests on $_SESSION['MESSAGE_AFTER_REDIRECT']

There are already some unit tests which check this variable;  the helper may be more convenient.

PR #432, opened 2018-04-13
Agent code refactor

### LINKED ISSUE #413: Refactor commands to devices
We can simplify the code in inc/agent.class.php

https://github.com/flyve-mdm/glpi-plugin/blob/develop/inc/agent.class.php#L509
checks for an undefined topic and send a message... there is nearly the same in sendPingQuery().

Same note for geolocation feature, but sendGeolocationQuery() can set a message with its first argument.

Refactor:
- use an exception in send*Query() to pass a message to the caller instead of a reference to a var (1st argument)
- the caller does not needs to check if the topic is not defined. The called method send*Query() will throw an exception if this happens

DIOHz0r:

### Changes description

Code refactoring for agent class

### Checklist

### References

Closes #413


inc/agent.class.php
```diff
@@ -1485,42 +1482,19 @@ public static function getTopicsToCleanup() {
...
-   }
-
-   /**
-    * Polls in the DB for a new geolocation entry with ID higher than the given one
-    * Timeouts if no new entry after a few seconds
-    * @param array $lastPosition
-    * @param string $errorMessage the error message to return to the caller
-    * @return boolean true if a new position found before timeout
-    */
-   protected function pollGeolocationAnswer($lastPosition, &$errorMessage) {
```
btry: If you drop the polling system (and I would agree with that) then the endpoint will behave differently. 

With the polling, the HTTP response code would be 200 only if a new position was recorded in DB.

Without the polling HTTP 200 will be send when the backend sucessfully send the geolocation request only.

If this PR is accepted, then we may need to update some documentation in the gh pages branch.

inc/agent.class.php
```diff
@@ -1530,38 +1504,26 @@ protected function pollGeolocationAnswer($lastPosition, &$errorMessage) {
...
-               $errorMessage = __('GPS is turned off or is not ready', 'flyvemdm');
-               return false;
-            } else {
-               return true;
+               throw new Exception(__('GPS is turned off or is not ready', 'flyvemdm'));
             }
+            return true;
          }
       }
+      throw new Exception(__('Timeout requesting position', 'flyvemdm'));
```
btry: same here: create an exception class

inc/agent.class.php
```diff
@@ -1575,36 +1537,23 @@ protected function pollInventoryAnswer() {
...
          $updatedInventory = array_pop($inventoryRows);
          if ($lastInventory === null && $updatedInventory !== null
-             || $lastInventory !== null && $lastInventory != $updatedInventory) {
+            || $lastInventory !== null && $lastInventory != $updatedInventory) {
             return true;
          }
       }
-
-      return false;
+      throw new Exception(__("Timeout querying the device inventory", 'flyvemdm'));
```
btry: same here

inc/agent.class.php
```diff
@@ -1615,8 +1564,7 @@ protected function pollPingAnswer() {
             return true;
          }
       }
-
-      return false;
+      throw new Exception(__("Timeout querying the device", 'flyvemdm'));
```
btry: again: create an exception class

PR #435, opened 2018-04-16
Refactor for unit and integration tests

DIOHz0r:

### Changes description

Refactor the tests for use common functions like file and package creation

### Checklist

### References

Closes #N/A


btry:
Hi

Globally OK for the refactor, but I think we can find a better name for createFlyvemdmDumbFile (and package).

Also I did a similar refactor in unit tests last week. For helper methods creating items, I think we should have one argument $input for the data of the item, like prepareInputForAdd().

PR #446, opened 2018-04-23
Getting an agent by topic failed

btry:
Fix getByTopic()

### DIFF
--- a/tests/suite-integration/PluginFlyvemdmAgent.php
+++ b/tests/suite-integration/PluginFlyvemdmAgent.php
@@ -943,4 +943,26 @@ class PluginFlyvemdmAgent extends CommonTestCase {
          $this->array($expected)->contains($aMessage['message']);
       }
    }
+
+   public function testGetByTopic() {
+      list($user, $serial, $guestEmail, $invitation) = $this->createUserInvitation(\User::getForeignKeyField());
+      $agent = $this->agentFromInvitation($user, $guestEmail, $serial,
+         $invitation->getField('invitation_token'));
+
+      // Test the agent is created
+      $this->boolean($agent->isNewItem())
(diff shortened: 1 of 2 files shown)


tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -943,4 +943,26 @@ private function compareListOfMQTTMessages(array $expected, array $received) {
...
+      $this->boolean($agent->isNewItem())
+         ->isFalse(json_encode($_SESSION['MESSAGE_AFTER_REDIRECT'], JSON_PRETTY_PRINT));
+
+      $computer = new \Computer();
+      $computer->getFromDB($agent->getField(\Computer::getForeignKeyField()));
+      $entityId = $computer->getField(\Entity::getForeignKeyField());
+      $serial = $computer->getField('serial');
+
+      $emptyAgent = new \PluginFlyvemdmAgent();
+      $emptyAgent->getByTopic("$entityId/agent/$serial");
```
DIOHz0r: Why is the serial used? this hasn't been implemented yet.

btry: PluginFlyvemdmAgent::getByTopic() finds the agent which matches a MQTT topic. the topic contains the serial of the associated computer. Then this test must use t he serial of the computer to test the method.

I don't remember in deep the creation of the agent in  the tests context, but you should generate a random string. If not, then this needs to be added now.

DIOHz0r: You're right however we should use the agent UUID and not the serial cause the serial could be duplicated and we will getting duplicated info also @rafaelje must change these topic value into the agent.

btry: I also have in mind the use of the UUID, but cor now the code uses the serial. We cannot replace the serial by the UUID a few days before a release.

Let's plan this for 2.1 or 3.0 depending on a potential BC.

PR #464, opened 2018-04-30
WIP Move to Circle CI

btry:
We need to parallelize all jobs.

Issues to solve
- php nightly build test

GLPI uses both Travis and Cricle CI. I think we should avoid it because we need to trigger a build (later) only when all jobs succeed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,70 @@
+version: 2
+
+shared: &shared
+  steps:
+    - checkout
+    - run:
+      name: M2M dependency
+      command: |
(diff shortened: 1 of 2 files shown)


btry:
After discussion with GLPI team to compare features between Circle CI and Travis CI, I found t hat Circle CI offers 4 parallel jobs, for the repo and its forks. Travis offers 5 parallel jobs per repo.

I applied a suggestion to disable coverage for most builds in #599 We reduced the linear build time from 3h30 to 1h30. This PR is now obsolete.

PR #466, opened 2018-05-01
Fix invited devices are always created in root entity

btry:
Automatically generate rules to "route" imported devices in the entity of  the invitation.

@rafaelje the tag in  the inventory must be similar to **invitation_4564546564587**.

I'm unsure how I described you the tag on telegram (underscore or dash). We need to check and use the same !

### DIFF
--- a/inc/fusioninventory.class.php
+++ b/inc/fusioninventory.class.php
@@ -1,0 +1,198 @@
+<?php
+/**
+ * LICENSE
+ *
+ * Copyright © 2016-2018 Teclib'
+ * Copyright © 2010-2018 by the FusionInventory Development Team.
+ *
+ * This file is part of Flyve MDM Plugin for GLPI.
+ *
+ * Flyve MDM Plugin for GLPI is a subproject of Flyve MDM. Flyve MDM is a mobile
+ * device management software.
+ *
+ * Flyve MDM Plugin for GLPI is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * Flyve MDM Plugin for GLPI is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


btry:
@DIOHz0r I'm planning to move later the code interacting with Fusion Inventory in the new file inc/fusioninventory.class.php. Especially portions of enrollment for import.

rafaelje:
I prefer to use "_" underscore

rafaelje:
This is a real example of the new information tag:

```
    <ACCOUNTINFO>
      <KEYNAME>TAG</KEYNAME>
      <KEYVALUE>invitation_b3aa766d6b62163799b6dbd6086a8281ec9ce32cac90da89e5e9968699ac623e</KEYVALUE>
    </ACCOUNTINFO>
```

rafaelje:
This is the PR related on the Android Agent:
https://github.com/flyve-mdm/android-mdm-agent/pull/405

btry:
Perfect I'll try today

inc/fusioninventory.class.php
```diff
@@ -0,0 +1,198 @@
...
+   /**
+    * Creates or updates an entity rule
+    *
+    * @param PluginFlyvemdmInvitation $invitation invitation to handle
+    */
+   public function addInvitationRule(PluginFlyvemdmInvitation $invitation) {
+      $entityId = $invitation->getField(Entity::getForeignKeyField());
+      $rule = $this->getRule($entityId);
+
+      $ruleCriteria = new RuleCriteria();
```
DIOHz0r: I have a question for this function

Don't you think it's better for performance to just compare if the rule exist and is active to then do a return?

If the rule is not active but exist, update it and if it not exist just add it

btry: A rule is combination of 3 different itemtypes:
- rule itself
- one or several criterias (with a boolean connector AND or OR)
- one or several actions

As I need to manupulate several itemtypes with relatios between them I choose to implement a class with this specific logic. The invitation class is already quite big.

I had 2 choices : 
- one rule per invitation with a single criteria
- one rule with a lot of criterias, one criteria per invitation

I choosed the 2nd, w th a boolean connector OR

PR #472, opened 2018-05-04
Update search options

btry:
Harmonize search options, and add missing entity / is_recursive for some itemtypes


btry:
@Gianfranco97 @hectorerb 

I removed in PluginFlyvemdmInvitation a search option to get the name of the invited user. This search option conflicts with the name of invitation I need to introduce to avoid some problems. 

This may impact some requests in the dashboard, then please, check the impact

PR #474, opened 2018-05-04
Add display preferences

btry:
Fix display preferences definition,

add display prefs for packages, and fix display of icon in list of packages

@Gianfranco97 I notify you, maybe you're interested about this UI ability. Note that the icon should be sent to you as raw data (png format)

Parts if code picked from Fdroid PR

### before
image

(and buggy display)
image

### after
image

### DIFF
--- a/inc/package.class.php
+++ b/inc/package.class.php
@@ -190,7 +190,7 @@ class PluginFlyvemdmPackage extends CommonDBTM {
       }
       try {
          $uploadedFile = $preparedFile['uploadedFile'];
-         $input['filename'] = 'flyvemdm/package/' . $this->fields['entities_id'] . '/' . uniqid() . '_' . basename($uploadedFile);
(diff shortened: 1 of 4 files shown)


inc/package.class.php
```diff
@@ -190,7 +190,7 @@ public function prepareInputForAdd($input) {
       }
       try {
          $uploadedFile = $preparedFile['uploadedFile'];
-         $input['filename'] = 'flyvemdm/package/' . $this->fields['entities_id'] . '/' . uniqid() . '_' . basename($uploadedFile);
+         $input['filename'] = 'flyvemdm/package/' . $input['entities_id'] . '/' . uniqid() . '_' . basename($uploadedFile);
```
btry: $this->fields should not contain anything or is garbage if an instance is r e-used.

PR #480, opened 2018-05-08
Update QR code size for easy scanning

### LINKED ISSUE #473: Improve QR code for invitations
**Is your feature request related to a problem? Please describe.**
The QR code has a lot of information to scan and it generates a condensed image with a large number of pixels making it difficult to scan specially when some devices hasn't have the property to focus their camera.

After a lot of tests I'd found that a minimal image size for solve this issue is 488px × 488px and the precision of  the code could be the lowest available.

**Describe the solution you'd like**
Make the image bigger on the invitation and the deeplink page to make the invitation more easy to scan. 

**Describe alternatives you've considered**
The other solution would be to reduce the quantity of information used on the QR code to generate less pixels

DIOHz0r:
### Changes description

Changed the QR code image to a bigger size so it could be more easy to scan it.

### References

Closes #473


btry:
I added a commit to update locales, and cleanup no longer localizable strings (the notification email messages are no longer translated, because the installer always uses english)

PR #490, opened 2018-05-10
Move and reverse unknown sources policy

btry:
The policy was in the bad category, and it is better to disallow unknown sources with the same mechanism as disable Bluetooth.

- if Disable Unknown Sources is applied with the value *Yes** then the agent must forbid the user from enabling unknown sources
- if Disable Unknown Sources is applied with the value **No**  then the agent must do nothing (allowing the user to freely enable or disable the unknown sources)
- if Disable Unknown Sources is not applied,  then the agent must do nothing  (allowing the user to freely enable or disable the unknown sources)

### DIFF
--- a/install/policies/deployment.php
+++ b/install/policies/deployment.php
@@ -101,4 +101,21 @@ return [
       'is_android_system'                   => '0',
       'is_apple_policy'                     => '0',
    ],
+
+   [
+      'name'                                => __('Disable unknown sources', 'flyvemdm'),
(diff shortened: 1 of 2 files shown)


btry:
/cc @rafaelje 

the PR contains information for your implementation

ajsb85:
typo in commit message: "nove" and reverse unknown sources policy

btry:
commit message fixed

PR #494, opened 2018-05-11
WIP Redesign use TLS policy

### LINKED ISSUE #414: Policy "Use TLS" enhancement
Several improvements:
- the policy must have a parameter to set the address and port of the MQTT broker for TLS
- the agent must try to connect with its credentials to check it can switch safely. If not, fallback to the previous MQTT broker with previous settings.

The policy must also allow (despite this is a bad idea) to switch from TLS to clear communication. To so so, if the agent gets the policy it must compare the settings with the current ones. If  they are different, it must apply them.

I set  this for 2.0 but maybe we should disable the policy for 2.0 and implement this stuff for 2.1 (release in progress)

btry:
Use TLS policy  is quite useless alone. This PR improves it to configure MQTT client of devices.

Scenarios 
- switch from clear MQTT to encrypted MQTT (or reverse)
- switch from a MQTT broker to an other (migration, horizontal scaling...)

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@diohz0r |1|

### References

Closes #414


btry:
related to #495

DIOHz0r:
I've rebased the code and solve a conflict, please update your local changes

btry:
done, thanks

PR #495, opened 2018-05-11
Remove not ready policy

btry:
This PR removes from the project the policies which are not ready for the release 2.0

/cc @DIOHz0r @rafaelje 

For convenience, please, remove yourself any policy that you want to remove, one per commit. Doing so will allow us to revert a commit to re-add the policy after the release.

### DIFF
--- a/install/policies/mdm.php
+++ b/install/policies/mdm.php
@@ -36,24 +36,6 @@ if (!defined('GLPI_ROOT')) {
 $category = 'Mobile Device Management';
 return [
    [
-      'name'                                => __('Use TLS', 'flyvemdm'),
-      'symbol'                              => 'useTLS',
-      'group'                               => 'MDM',
-      'type'                                => 'bool',
-      'type_data'                           => '',
-      'unicity'                             => 1,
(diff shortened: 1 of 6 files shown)


DIOHz0r:
I have done the rebase and update for this PR, please review it

ajsb85:
I don't agree, we can hide the work in progress instead of remove it.
it's a double of work.

btry:
@ajsb85 

I had also the idea of a new column to disable policies but : 
- users may find it and try it by resetting the disable bool
- we need to add and maintain upgrade code

removing the policies in the install code is easy, and revertable with ```git revert```. This is way less work

DIOHz0r:
Another reason is the last two commits, both policies removed can't be applied on Android systems according to @rafaelje

btry:
when all policies to removed are in this PR we will need an extra commit to remove them when upgrading: we have some users running the development branch.

DIOHz0r:
> when all policies to removed are in this PR we will need an extra commit to remove them when upgrading: we have some users running the development branch.

Actually this PR is ready IMO.

PR #502, opened 2018-05-15
Invalid fleet and task relation 

DIOHz0r:
### Changes description

Upgrade script for dev throws a problem adding an index for a not existing column also the relation between the Fleets and Taks are broken for search options, this PR fixes the problem.

### References

imagen
imagen


btry:
@DIOHz0r in your Postman screenshot, the requested itemtype is not specified. I read ```:itemtype```

I guess this was ```PluginFlyvemdmTask``` ?

In one of my commits to this PR, I had to udpate the search options (as you told me, there was a problem with them). However this may impact the way to get items from the API. What is the context and the purpose of the request ?

DIOHz0r:
@btry , the problem is when you try to get the fleet forcing to display cols 3 or 4 (both are task).

btry:
The columns 3 and 4 should be handled by CommonDBRelation. Have a look on its implementation of getSearchOptionsNew(). There is an implementation issue I fixed with a commit in this PR. Unfortunately, this makes less search options available for an API consumer, and we need to find an other way to get the informations you tried to collect with your example API request.

install/mysql/plugin_flyvemdm_empty.sql
```diff
@@ -201,15 +201,17 @@ CREATE TABLE `glpi_plugin_flyvemdm_policies` (
 DROP TABLE IF EXISTS `glpi_plugin_flyvemdm_tasks`;
 CREATE TABLE `glpi_plugin_flyvemdm_tasks` (
   `id`                                         int(11)      NOT NULL AUTO_INCREMENT,
-  `itemtype_applied`                           varchar(255) DEFAULT NULL,
-  `items_id_applied`                           int(11)      NOT NULL DEFAULT '0',
+  `policies_itemtype`                          varchar(255) DEFAULT NULL,
+  `policies_items_id`                          int(11)      NOT NULL DEFAULT '0',
```
btry: this change breaks the implementation in CommonDBRelation ! Revert it, please

install/mysql/plugin_flyvemdm_empty.sql
```diff
@@ -201,15 +201,17 @@ CREATE TABLE `glpi_plugin_flyvemdm_policies` (
...
   `plugin_flyvemdm_policies_id`                int(11)      NOT NULL DEFAULT '0',
   `value`                                      varchar(255) NOT NULL DEFAULT '',
   `itemtype`                                   varchar(255) DEFAULT NULL,
   `items_id`                                   int(11)      NOT NULL DEFAULT '0',
   PRIMARY KEY (`id`),
-  KEY `FK_applied` (`items_id_applied`, `itemtype_applied`),
-  KEY `plugin_flyvemdm_policies_id` (`plugin_flyvemdm_policies_id`)
+  KEY `plugin_flyvemdm_policies_id` (`plugin_flyvemdm_policies_id`),
+  KEY `FK_policies_applied` (`policies_items_id`, `policies_itemtype`),
```
btry: The KEY `items_id_applied`, `itemtype_applied` must be in the same key

setup.php
```diff
@@ -99,7 +99,7 @@ function plugin_init_flyvemdm() {
       plugin_flyvemdm_registerClasses();
       plugin_flyvemdm_addHooks();
 
-      $CFG_GLPI['fleet_types'] = [PluginFlyvemdmFile::class, PluginFlyvemdmPackage::class];
+      $CFG_GLPI['policy_use_types'] = [PluginFlyvemdmFleet::class, PluginFlyvemdmAgent::class];
```
btry: OK to change the name of this key, but it is not conform. I think ```notifiable_types``` is better

DIOHz0r:
Merged becuase @hectorerb needs these changes and it was worked and reviewed with @btry

btry:
We need to fix the call to migrateonetable !

PR #515, opened 2018-05-21
New fields to display on search

### LINKED ISSUE #513: Allow display all fields from PluginFlyvemdmTask
**Is your feature request related to a problem? Please describe.**
When request a searchItems to PluginFlyvemdmTask does not allow to show all the fields in this table

for example to request a listSearchOptions to PluginFlyvemdmTask fields 6 and 7 are set to not show, the uid field shows that these fields are not obtained from the relationship with other tables

captura de pantalla 2018-05-21 a la s 10 54 58

**Describe the solution you'd like**
Set in listSearchOptions value nodisplay on false to fields 6 and 7 or all field that don't come from a relationship with another table

DIOHz0r:
### Changes description

Add the option to display new task fields on search results.

### References

Closes #513

### DIFF
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -534,7 +534,6 @@ class PluginFlyvemdmTask extends CommonDBRelation {
          'field'         => 'itemtype',
          'name'          => __('itemtype'),
          'massiveaction' => false,
-         'nodisplay'     => '1',
          'datatype'      => 'string',
       ];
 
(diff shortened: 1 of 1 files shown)


btry:
It seems this issue shows that some search options should be hidden in GLPU UI but available to an API consumer. We probably have to discuss about a new search options feature in GLPI.

Issue #503, opened 2018-05-16
Code refactor needed


btry:
The same code appears in 2 different places (see #502 , the foreach loop)
- PluginFlyvemdmFile::post_updateItem()
- PluginFlyvemdmPackage::post_updateItem()

We need to create 
- an abstract class, parent of both classes named ```PluginFlyvemdmDeployable```
- which implements an interface named ```PluginFlyvemdmDeployableInterface```

Maybe some more code may be mutualized.

/cc @Naylin15 if she want to do it.

DIOHz0r:
I think we could close this issue because the were a lot of refactor and the deployable class was created.

PR #524, opened 2018-05-22
Code refactor

DIOHz0r:
### Changes description

Refactor and clean up of current project code 

### References

Ref. #503

### DIFF
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -773,7 +773,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
 
       $tab[] = [
          'id'                 => '3',
-         'table'              => 'glpi_plugin_flyvemdm_fleets',
+         'table'              => PluginFlyvemdmFleet::getTable(),
          'field'              => 'name',
          'name'               => __('Fleet', 'flyvemdm'),
          'datatype'           => 'dropdown'
@@ -781,7 +781,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
 
       $tab[] = [
          'id'                 => '4',
-         'table'              => 'glpi_computers',
+         'table'              => Computer::getTable(),
          'field'              => 'id',
          'name'               => __('Computer'),
          'datatype'           => 'dropdown',
@@ -790,7 +790,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
 
       $tab[] = [
          'id'                 => '7',
-         'table'              => 'glpi_plugin_flyvemdm_fleets',
+         'table'              => PluginFlyvemdmFleet::getTable(),
          'field'              => 'id',
          'name'               => __('Fleet - ID'),
          'massiveaction'      => false,
@@ -871,7 +871,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
 
       $tab[] = [
          'id'                 => '17',
-         'table'              => 'glpi_entities',
+         'table'              => Entity::getTable(),
          'field'              => 'completename',
          'name'               => __('Entity'),
          'datatype'           => 'dropdown'
@@ -879,7 +879,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
 
       $tab[] = [
          'id'                 => '18',
-         'table'              => 'glpi_plugin_flyvemdm_policies',
+         'table'              => PluginFlyvemdmPolicy::getTable(),
          'field'              => 'name',
          'name'               => __('Applied policy', 'flyvemdm'),
          'datatype'           => 'dropdown',
@@ -887,7 +887,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
          'nosort'             => true,
          'joinparams'         => [
             'beforejoin'         => [
-               'table'           => 'glpi_plugin_flyvemdm_tasks',
+               'table'           => PluginFlyvemdmTask::getTable(),
                'joinparams'      => [
                   'jointype'     => 'child',
                   'linkfield'    => 'items_id_applied',
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -1346,8 +1346,13 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
       // @see User::post_addItem
       $profileId = $config['agent_profiles_id'];
       $agentUserId = $agentAccount->getID();
-      $DB->query("UPDATE `glpi_profiles_users` SET `profiles_id` = '$profileId'
-                  WHERE `users_id` = '$agentUserId'");
+      try {
+         $DB->query("UPDATE `glpi_profiles_users` SET `profiles_id` = '$profileId'
+                     WHERE `users_id` = '$agentUserId'");
+      } catch (GlpitestSQLError $e) {
+         Toolbox::logInFile('php-errors',
+            "Plugin Flyvemdm : Could not update profile id = '$profileId'");
+      }
 
       $agentToken = User::getToken($agentAccount->getID(), 'api_token');
       if ($agentToken === false) {
@@ -1780,7 +1785,12 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
                ];
 
                $mqttUser = new PluginFlyvemdmMqttuser();
-               $mqttClearPassword = PluginFlyvemdmMqttuser::getRandomPassword();
+               try {
+                  $mqttClearPassword = PluginFlyvemdmMqttuser::getRandomPassword();
+               } catch (Exception $e) {
+                  Session::addMessageAfterRedirect($e->getMessage(), true, ERROR);
+                  return;
+               }
                if (!$mqttUser->getByUser($serial)) {
                   // The user does not exists
                   $mqttUser->add([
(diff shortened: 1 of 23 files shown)


inc/deployable.class.php
```diff
@@ -0,0 +1,188 @@
...
+ * @link      https://flyve-mdm.com/
+ * ------------------------------------------------------------------------------
+ */
+
+class PluginFlyvemdmDeployable extends CommonDBTM {
+
+   /**
+    * @var bool $usenotepad enable notepad for the itemtype (GLPi >=0.85)
+    */
+   protected $usenotepadRights = true;
```
btry: not related to a deployable object. I think this is better to keep it in the final class.

inc/deployable.class.php
```diff
@@ -0,0 +1,188 @@
...
+   /**
+    * @var bool $usenotepad enable notepad for the itemtype (GLPi < 0.85)
+    */
+   protected $usenotepad = true;
```
btry: Same comment as above. By the way, these 2 variables are duplicated for compatibility reasons across 2 versions of GLPI (the name of the property changed). If I remember correctly this was somewhere in 0.90.x releases, and we are no longer compatible with them.

inc/deployable.class.php
```diff
@@ -0,0 +1,188 @@
...
+
+
+   /**
+    * Sends a file
+    */
+   protected function sendFile(array $fields) {
```
btry: Can you convert $fields argument into several arguments for readability ?
(and update docblock)

inc/deployable.class.php
```diff
@@ -0,0 +1,188 @@
...
+      // Endnow to prevent any unwanted bytes
+      exit(0);
+   }
+
+   /**
+    * @param string $itemtype
+    * @param integer $itemId
+    * @param PluginFlyvemdmTask $task
+    */
+   protected function publishTaskAction($itemtype, $itemId, PluginFlyvemdmTask $task) {
```
btry: I suggest to rename this method post_updateItem() and reuse the calls to PluginFlyvemdmTask::publishPolicy() and  PluginFlyvemdmTask::createTaskStatuses() for  readability.

In the 2 inherited classes, just call parent::post_updateItem()

inc/file.class.php
```diff
@@ -299,119 +274,13 @@ public function post_updateItem($history = 1) {
       $itemtype = $this->getType();
       $itemId = $this->getID();
```
btry: These 2 lines are in File and Package, then you can move them in  the common implementation of publishTaskAction() (which I suggested to rename into post_updateItem())

inc/deployable.class.php
```diff
@@ -0,0 +1,188 @@
...
+    * Create a directory
+    * @param string $dir
+    */
+   protected function createEntityDirectory($dir) {
+      if (!is_dir($dir)) {
+         @mkdir($dir, 0770, true);
+      }
+   }
+
+   protected function purgeItem($filename) {
```
btry: the name of the method may lead to confusion with purge (meaning deleting am item from database without putting it in the trash). We should find a better name which means we remove a file from the file system.

inc/file.class.php
```diff
@@ -36,20 +36,12 @@
 /**
  * @since 0.1.30
  */
-class PluginFlyvemdmFile extends CommonDBTM {
-
-   // name of the right in DB
-   static $rightname = 'flyvemdm:file';
-
-   /**
-    * @var bool $usenotepad enable notepad for the itemtype (GLPi < 0.85)
```
btry: here and a few lines below you have the exact informations I talked about in the beginning of the review. The variable for GLPI < 0.85 may be safely dropped.

inc/file.class.php
```diff
@@ -198,24 +190,7 @@ private function getUploadedFile($uploadMandatory = false) {
...
-
-   /**
-    * Create a directory
-    * @param string $dir
-    */
-   protected function createEntityDirectory($dir) {
-      if (!is_dir($dir)) {
-         @mkdir($dir, 0770, true);
-      }
+      return $this->getDeployableURL('file');
```
btry: Maybe we can replace calls to getFileURL() by getDeployableURL() directly (and do ther same in package).

This way, getDeployableURL() may be shared in the Deployable class.

inc/deployable.class.php
```diff
@@ -0,0 +1,188 @@
...
+ * ------------------------------------------------------------------------------
+ * @author    Thierry Bugier
+ * @copyright Copyright © 2018 Teclib
+ * @license   AGPLv3+ http://www.gnu.org/licenses/agpl.txt
+ * @link      https://github.com/flyve-mdm/glpi-plugin
+ * @link      https://flyve-mdm.com/
+ * ------------------------------------------------------------------------------
+ */
+
+class PluginFlyvemdmDeployable extends CommonDBTM {
```
btry: this class should be abstract and implement an interface PluginFlyvemdmDeployableInterface

DIOHz0r: I was doing in that way but I didn't see the meaning of set it as an abstract class with interface, most of the methods implemented are coming from CommonDBTM. Could you give some examples?

inc/deployable.class.php
DIOHz0r: actually file and application class had the property and same value.

inc/deployable.class.php
DIOHz0r: same previous comment for this

inc/deployable.class.php
DIOHz0r: post_updateItem must be compatible with its parent method on commonDBTM so I can't use that name with those arguments also the child classes have different logic implementations. 

This method now is a little more isolated and testable.

inc/file.class.php
DIOHz0r: These lines aren't on the Package class.

inc/task.class.php
```diff
@@ -410,6 +417,7 @@ public function createTaskStatuses(PluginFlyvemdmNotifiableInterface $item) {
       }
 
       // Initialize a task status for each agent in the fleet
+      $rows = [];
```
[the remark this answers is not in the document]
DIOHz0r: if both conditionals fails then the foreach action will give a PHP error code because the $row is not defined as array

inc/agent.class.php
```diff
@@ -1781,7 +1786,12 @@ protected function setupMqttAccess() {
                ];
 
                $mqttUser = new PluginFlyvemdmMqttuser();
-               $mqttClearPassword = PluginFlyvemdmMqttuser::getRandomPassword();
+               try {
+                  $mqttClearPassword = PluginFlyvemdmMqttuser::getRandomPassword();
+               } catch (Exception $e) {
+                  Session::addMessageAfterRedirect($e->getMessage(), true, ERROR);
+                  return;
+               }
```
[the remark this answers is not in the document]
DIOHz0r: This case is not an test, it is a real throw action by the random password method which is not catch.

tests/suite-integration/PluginFlyvemdmTask.php
```diff
@@ -221,25 +165,45 @@ public function testApplyPolicyOnFleet() {
...
+      // Check task statuses are deleted
+      $rows = $taskStatus->find("`$taskFk` = '$taskId'");
+      $this->array($rows)->size->isEqualTo(0);
+   }
+
+   /**
+    * @param \PluginFlyvemdmNotifiableInterface $notifiableItem
+    * @return array
+    */
+   private function checkNotifiableMqttMessage($notifiableItem) {
```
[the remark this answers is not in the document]
DIOHz0r: Agent or Fleet, you set that name ;)

inc/fleet.class.php
```diff
@@ -533,7 +534,11 @@ public function refreshPersistedNotifications() {
       ];
       foreach ($DB->request($request) as $row) {
          $task->getFromDB($row['id']);
-         $task->publishPolicy($this);
+         try {
+            $task->publishPolicy($this);
+         } catch (TaskPublishPolicyPolicyNotFoundException $exception) {
+            // we shall not send technical information to the error message.
+         }
```
btry: PHP CS will complain. If you want to send a message to the end-user, use a user friendly message. If you want to keep a log, feel free to use Toolbox::logInFile()

inc/fleet.class.php
```diff
@@ -339,11 +342,10 @@ public function cleanDBonPurge() {
...
 
       // Unsuscribe all agents from the fleet
+      $result = false;
       $fleetId = $this->getID();
-      $query = "SELECT `id`
-      FROM `glpi_plugin_flyvemdm_agents`
-      WHERE `glpi_plugin_flyvemdm_agents`.`plugin_flyvemdm_fleets_id` = '$fleetId'";
-
+      $agentTable = PluginFlyvemdmAgent::getTable();
+      $query = "SELECT `id` FROM `$agentTable` WHERE `$agentTable`.`plugin_flyvemdm_fleets_id` = '$fleetId'";
```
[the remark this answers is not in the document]
DIOHz0r: better let's do it for another PR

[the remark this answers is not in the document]
btry: OK for me, then revert the change here and create an other PR

DIOHz0r:
@btry, do you have any new review for this?

btry:
Hi,

I was waiting for updates to solve this: 
https://github.com/flyve-mdm/glpi-plugin/pull/524#discussion_r190487010

tests/suite-integration/PluginFlyvemdmTask.php
```diff
@@ -274,28 +238,25 @@ public function testApplyPolicyOnAgent() {
...
-      $rows = $log->find("`id` > '$mqttLogId' AND `topic` = '$expectedTopic'");
-      $this->array($rows)->size->isEqualTo(1);
-      $row = array_pop($rows);
-
-      // Check the message
-      $this->string($row['message'])->isEqualTo('');
-
-      // Check task statuses are deleted
-      $rows = $taskStatus->find("`$taskFk` = '$taskId'");
-      $this->array($rows)->size->isEqualTo(0);
```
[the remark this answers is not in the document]
btry: OK, thank you. I'll review this specific part and feedback. As I'm currently busy, you may need to wait friday.

PR #509, opened 2018-05-17
Add min and max version of OS for policies

Naylin15:
### Changes description
Add min and max version for policy according the operating system

### Checklist

Please check if your PR fulfills the following specifications:

- [ ] Tests for the changes have been added
- [x] Docs have been added/updated

### References

Closes #443
related: https://github.com/flyve-mdm/web-mdm-dashboard/issues/740

PR #527, opened 2018-05-23
Patch for PR of OS min/max versions

DIOHz0r:
### Changes description
Add missed code for the PR #509

### DIFF
--- a/install/upgrade/update_to_dev.php
+++ b/install/upgrade/update_to_dev.php
@@ -157,10 +157,10 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
    $table = 'glpi_plugin_flyvemdm_policies';
    $migration->addField($table, 'recommended_value', 'string', ['after' => 'default_value']);
    $migration->addField($table, 'is_android_system', 'bool', ['after' => 'recommended_value']);
-   $migration->addField($table, 'android_min_version', 'bool', ['after' => 'is_android_system']);
-   $migration->addField($table, 'android_max_version', 'bool', ['after' => 'android_min_version']);
-   $migration->addField($table, 'apple_min_version', 'bool', ['after' => 'android_max_version']);
-   $migration->addField($table, 'apple_max_version', 'bool', ['after' => 'apple_min_version']);
+   $migration->addField($table, 'android_min_version', 'string', ['after' => 'is_android_system', 'value' => '0']);
+   $migration->addField($table, 'android_max_version', 'string', ['after' => 'android_min_version', 'value' => '0']);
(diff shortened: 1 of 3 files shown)


install/policies/mdm.php
```diff
@@ -43,7 +43,7 @@
       'type_data'                           => '',
       'unicity'                             => 1,
       'plugin_flyvemdm_policycategories_id' => $category,
-      'comment'                             => __('Use TLS.', 'flyvemdm'),
+      'comment'                             => __('Use TLS', 'flyvemdm').'.',
```
btry: put the dot in the localizable string

install/upgrade/update_to_dev.php
```diff
@@ -163,6 +163,12 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
    $migration->addField($table, 'apple_max_version', 'bool', ['after' => 'apple_min_version']);
    $migration->addKey($table, 'group', 'group');
    $migration->addKey($table, 'plugin_flyvemdm_policycategories_id', 'plugin_flyvemdm_policycategories_id');
+   if ($DB->fieldExists($table, 'is_android_policy')) {
+      $migration->dropField($table, 'is_android_policy');
+   }
+   if ($DB->fieldExists($table, 'is_apple_policy')) {
+      $migration->dropField($table, 'is_apple_policy');
+   }
```
btry: Migration::dropField() already checks if the field exists before actually dropping it. This change is useless

install/policies/mdm.php
DIOHz0r: Why? the dot it's at the end and we could use the same translation of the upper line or do you really want to have to translate two string just for only one dot?

ajsb85:
Hi, @DIOHz0r 
`fix(install): fix for db schema update and locale`
Should be:
`fix(install): db schema update and locale`
Thanks

Naylin15:
Fixing issues that complement this PR and making several tests in the server

I think I found the solution but I need to test it some more

One of the main issues was that the DB was not being updated properly, the values of the os versions were not being added when running cli_install.php

btry:
Indeed. I'll point to you the code that you need to update.

btry:
@Naylin15 have a look into the PluginFlyvemdmInstaller::createPolicies()

This method is in charge of creating / updating the policies in the database. You need to modify it to take into account the new columns you introduced. Once done, the update process should run as you expected.

install/policies/mdm.php
btry: Maybe in some languages the punctuation may be very different to latin languages. We cannot do assumption about locales.

Moreover this policy willl change its name see #494 We wil need to improve comment and probably use longer text than only copying the name.

PR #534, opened 2018-05-28
Fix agent and fleet notification on upgrade

DIOHz0r:
### Changes description

A bug related with the search for the removed policies is finding only the fleets, the agents aren't notified so this PR solves it. 

This change also solve another performance problem sending only one notification through the broker.

raw query for you review:
``` SQL
SELECT
  `glpi_plugin_flyvemdm_policies`.`symbol`,
  `glpi_plugin_flyvemdm_tasks`.`id`,
  `glpi_plugin_flyvemdm_tasks`.`itemtype_applied`,
  `glpi_plugin_flyvemdm_tasks`.`items_id_applied`,
  `glpi_plugin_flyvemdm_fleets`.`entities_id`
FROM `glpi_plugin_flyvemdm_policies`, `glpi_plugin_flyvemdm_tasks`, `glpi_plugin_flyvemdm_fleets`,
  `glpi_plugin_flyvemdm_agents`
WHERE `glpi_plugin_flyvemdm_tasks`.`plugin_flyvemdm_policies_id` = `glpi_plugin_flyvemdm_policies`.`id`
      AND `symbol` IN
          ('disableFmRadio', 'disableVoiceMail', 'disableCallAutoAnswer', 'disableVoiceDictation', 'disableUsbOnTheGo', 
           'resetPassword', 'inventoryFrequency', 'disableSmsMms', 'disableStreamVoiceCall', 'disableCreateVpnProfiles')
      AND ((((`itemtype_applied` = 'PluginFlyvemdmFleet' AND
              `glpi_plugin_flyvemdm_tasks`.`items_id_applied` = `glpi_plugin_flyvemdm_fleets`.`id`)) OR
            ((`itemtype_applied` = 'PluginFlyvemdmAgent' AND
              `glpi_plugin_flyvemdm_tasks`.`items_id_applied` = `glpi_plugin_flyvemdm_agents`.`id`))))
GROUP BY `id`
```

### DIFF
--- a/install/upgrade/update_to_dev.php
+++ b/install/upgrade/update_to_dev.php
@@ -199,28 +199,41 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
    ];
    $tasksTable = 'glpi_plugin_flyvemdm_tasks';
    $fleetTable = 'glpi_plugin_flyvemdm_fleets';
+   $agentsTable = 'glpi_plugin_flyvemdm_agents';
    $request = [
       'FIELDS'     => [
-         $table => ['symbol'],
-         $tasksTable => ['id', 'itemtype_applied', 'items_id_applied'],
-         $fleetTable => ['entities_id'],
+         $table       => ['symbol'],
+         $tasksTable  => ['id', 'itemtype_applied', 'items_id_applied'],
+         $fleetTable  => ['entities_id'],
       ],
       'FROM'       => $table,
       'INNER JOIN' => [
          $tasksTable => [
             'FKEY' => [
                $tasksTable => 'plugin_flyvemdm_policies_id',
-               $table => 'id'
-            ]
+               $table      => 'id',
+            ],
          ],
-         $fleetTable => [
-            'FKEY' => [
+      ],
+      'LEFT JOIN'  => [
+         $fleetTable  => [
+            'itemtype_applied' => 'PluginFlyvemdmFleet',
+            'FKEY'             => [
                $tasksTable => 'items_id_applied',
                $fleetTable => 'id',
             ],
          ],
+         $agentsTable => [
+            'itemtype_applied' => 'PluginFlyvemdmAgent',
+            'FKEY'             => [
+               $tasksTable  => 'items_id_applied',
+               $agentsTable => 'id',
+            ],
+         ],
       ],
-      'WHERE'      => ['symbol' => $policies],
+      'WHERE'      => [
+         'symbol' => $policies,
+      ],
    ];
    $result = $DB->request($request);
    if (count($result) > 0) {
@@ -229,10 +242,14 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
          switch ($data['itemtype_applied']) {
             case PluginFlyvemdmFleet::class:
                $type = 'fleet';
+               $entityId = $data['entities_id'];
                break;
 
             case PluginFlyvemdmAgent::class:
+               $agent = new PluginFlyvemdmAgent();
+               $agent->getFromDB($data['items_id_applied']);
                $type = 'agent';
+               $entityId = $agent->getEntityID();
                break;
 
             default:
@@ -243,7 +260,7 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
          }
          $topic = implode('/', [
             '/',
-            $data['entities_id'],
+            $entityId,
             $type,
             $data['items_id_applied'],
             'Policy',


ajsb85:
HI, @diohz0r 
The commit message is not clear:
`update code for upgrade installation`
In spanish is also not clear and redundant:
`código de actualización para instalación de actualización`
And please don't paste pictures of the code, paste the code block with the proper syntax highlighting.
Thanks

btry:
Moving the join condition in the where clause is OK, but ad far as I know, FKEY is not recognized here. I think this explains your use of group by: this workarounds the malfunction of the join. I'll suggest to youban alternative in a few hours

DIOHz0r:
> After reading the generated query, it sems you're lucky the FKEY key in WHERE clause does not do anything :D. 

It does.. they are on the where criteria (`glpi_plugin_flyvemdm_tasks`.`items_id_applied` = `glpi_plugin_flyvemdm_fleets`.`id`)

> We have a JOIN between tasks and policies, then why not use a INNER JOIN clause instead of limiting rows in the WHERE clause ? This would make the query more readable.

The inner join needs two conditions to work, the itemtype_applied must match with the correct table and the items_id_applied with the id on that table, I didn't find how to do it with the query builder.

Also I tested the query on dev and it gets the same tasks repeated a lot of times, I tried using a select distinct but the builder never added the distinct word.

DIOHz0r:
I've updated the code to get the following query
```SQL
EXPLAIN
SELECT
 `glpi_plugin_flyvemdm_policies`.`symbol`,
 `glpi_plugin_flyvemdm_tasks`.`id`,
 `glpi_plugin_flyvemdm_tasks`.`itemtype_applied`,
 `glpi_plugin_flyvemdm_tasks`
 .`items_id_applied`,
 `glpi_plugin_flyvemdm_fleets`.`entities_id`
FROM `glpi_plugin_flyvemdm_policies`
INNER JOIN `glpi_plugin_flyvemdm_tasks` ON (`glpi_plugin_flyvemdm_tasks`.`plugin_flyvemdm_policies_id` = `glpi_plugin_flyvemdm_policies`.`id`)
LEFT JOIN `glpi_plugin_flyvemdm_fleets` ON ((`itemtype_applied` = 'PluginFlyvemdmFleet' AND `glpi_plugin_flyvemdm_tasks`.`items_id_applied` =
 `glpi_plugin_flyvemdm_fleets`.`id`))
LEFT JOIN `glpi_plugin_flyvemdm_agents` ON (`itemtype_applied` = 'PluginFlyvemdmAgent' AND
 `glpi_plugin_flyvemdm_tasks`.`items_id_applied` =
 `glpi_plugin_flyvemdm_agents`.`id`)
WHERE `symbol` IN
 ('disableFmRadio', 'disableVoiceMail', 'disableCallAutoAnswer', 'disableVoiceDictation', 'disableUsbOnTheGo', 
 'resetPassword', 'inventoryFrequency', 'disableSmsMms', 'disableStreamVoiceCall', 'disableCreateVpnProfiles')
```

install/upgrade/update_to_dev.php
```diff
@@ -210,17 +211,29 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
...
-            ]
+               $table      => 'id',
+            ],
          ],
-         $fleetTable => [
-            'FKEY' => [
+      ],
+      'LEFT JOIN'  => [
+         $fleetTable  => [
+            'itemtype_applied' => 'PluginFlyvemdmFleet',
```
btry: If I remember correctly, this use case if JOIN is not documented. I assumed this was not possible. If this works, maybe we should add a contribution to the developer documentation.

btry: The request is exacly how I wanted it to be. Nice.

Remains to find how to get the entities_id column from an agent without conflict with the entities_id from a fleet. Again, I remember that column aliasing is not supported by the query builder.

Getting the entiteis_id column from agents is needed to build the correct MQTT topic in the foreach() :/

DIOHz0r: I added a new condition to agent join now the new query is this....

```SQL
SELECT
  `glpi_plugin_flyvemdm_policies`.`symbol`,
  `glpi_plugin_flyvemdm_tasks`.`id`,
  `glpi_plugin_flyvemdm_tasks`.`itemtype_applied`,
  `glpi_plugin_flyvemdm_tasks`.`items_id_applied`,
  `glpi_plugin_flyvemdm_fleets`.`entities_id`
FROM `glpi_plugin_flyvemdm_policies`
  INNER JOIN `glpi_plugin_flyvemdm_tasks`
    ON (`glpi_plugin_flyvemdm_tasks`.`plugin_flyvemdm_policies_id` = `glpi_plugin_flyvemdm_policies`.`id`)
  LEFT JOIN `glpi_plugin_flyvemdm_fleets` ON (`itemtype_applied` = 'PluginFlyvemdmFleet' AND
                                              `glpi_plugin_flyvemdm_tasks`.`items_id_applied` =
                                              `glpi_plugin_flyvemdm_fleets`.`id`)
  LEFT JOIN `glpi_plugin_flyvemdm_agents` ON (`itemtype_applied` = 'PluginFlyvemdmAgent' AND
                                              `glpi_plugin_flyvemdm_tasks`.`items_id_applied` =
                                              `glpi_plugin_flyvemdm_agents`.`id` AND
                                              (`glpi_plugin_flyvemdm_agents`.`plugin_flyvemdm_fleets_id` =
                                               `glpi_plugin_flyvemdm_fleets`.`id`))
WHERE `symbol` IN
      ('disableFmRadio', 'disableVoiceMail', 'disableCallAutoAnswer', 'disableVoiceDictation', 'disableUsbOnTheGo',
       'resetPassword', 'inventoryFrequency', 'disableSmsMms', 'disableStreamVoiceCall', 'disableCreateVpnProfiles')
```

btry:
Hi

I disagree with the latest stage of the query.  

For agents we must cet thé entities_id from agents table, and for fleets we must cet entities_id from fleets table. Consider that an agent may be in a different entity than the fleet (is_recursive column).

If you cannot find a solution I'm a single query (this may be the case) then split the request into 2 requests.

btry:
Hi

I disagree with the latest stage of the query.  

For agents we must cet thé entities_id from agents table, and for fleets we must cet entities_id from fleets table. Consider that an agent may be in a different entity than the fleet (is_recursive column).

If you cannot find a solution I'm a single query (this may be the case) then split the request into 2 requests.

DIOHz0r:
I've updated again the query to do another search when the itemtype applied is an Agent.

install/upgrade/update_to_dev.php
```diff
@@ -199,51 +199,67 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
...
+         'symbol' => $policies,
       ],
-      'WHERE'      => ['symbol' => $policies],
    ];
    $result = $DB->request($request);
    if (count($result) > 0) {
       $mqttClient = PluginFlyvemdmMqttclient::getInstance();
       foreach ($result as $data) {
+         $type = '';
+         $entities_id = 0;
```
btry: dangerous: it may send a mqtt message if  the row does not match none of fleet and agent itemtypes.

Add a ```default``` case in  the switch, with a ```continue``` instruction.

install/upgrade/update_to_dev.php
```diff
@@ -199,51 +199,67 @@ function plugin_flyvemdm_update_to_dev(Migration $migration) {
...
          switch ($data['itemtype_applied']) {
             case PluginFlyvemdmFleet::class:
                $type = 'fleet';
+               $entities_id = $data['entities_id'];
```
btry: I prefer $entityId for the name of the variable. I'm used to this in the whole project (and consistency matters)

PR #551, opened 2018-06-13
Type of is_online is specific

btry:
Probably a loss due to a rebase. A fix on search options was created / merged nearly at the same  time as #474


ajsb85:
Hi, @btry 
But what we need to show?
Because bool is used for checkbox layout.

btry:
Hi

the following pic comes from #474 
https://user-images.githubusercontent.com/14139801/39638182-acd24726-4fc5-11e8-99bf-9da7bfa73b6e.png

I found that the light bulbs are not displayed. Istead we have "Yes" and "No". The UI enhancement is broken, and this PR fixes it.

btry:
If bool is used by the dashboard, then we need to do some kind of override like we do in GLPI or plugins.

See getSpecificValueToDisplay() methods in GLPI or many plugins, Flyve MDM included.

PR #552, opened 2018-06-13
New code refactor

DIOHz0r:
### Changes description

New code refactor for the source code of the plugin

### References

Ref #503

### DIFF
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -310,11 +310,7 @@ class PluginFlyvemdmAgent extends CommonDBTM implements PluginFlyvemdmNotifiable
          return false;
       }
 
-      if (isset($_GET["start"])) {
-         $start = intval($_GET["start"]);
-      } else {
-         $start = 0;
-      }
+      $start = isset($_GET["start"]) ? intval($_GET["start"]) : 0;
 
       $dbUtils = new DbUtils();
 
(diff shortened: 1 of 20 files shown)


inc/agent.class.php
```diff
@@ -488,12 +487,12 @@ public function prepareInputForUpdate($input) {
 
       // send wipe to the agent
       if (isset($input['wipe']) && $input['wipe'] != '0') {
-         $input['wipe'] == '1';
```
btry: What ? I wrote that ?

btry:
I don't like much the style of affecting a variable and re-affect in a if statement.It causes 2 or 3 instructions. 

The style I adopted causes 1 or 2 instructions execution. 
Maybe should we search or benchmark the performance of both

DIOHz0r:
Hi,

The idea of that change is for better code reading according to bets practices. We can also simply some of those if using the following syntax 

```php 
$var = ($a && $b)? $a : 0;
```

btry:
I would agree to use the ternary operator. But we need to keep the condition easy to read. I took the habit to split it into 3 lines if it is complex

(Condition)
? Alternative1
: Alternative 2;

inc/agent.class.php
DIOHz0r: yes you did and it has been there for a lot of time

PR #553, opened 2018-06-15
Notifiable agents refactor

### LINKED ISSUE #528: Code refactor : intermediate abstract classe between CommonDBTM and (Fleet / Agent)
We have a NotifiableInterface, but not yet a Notifiable class

Roughly : 

```php
Class PluginFlyvemdmNotifiable implements PluginFlyvemdmNotifiableInterface {
   // ...
}


class PluginFlyvemdmAgent extends PluginFlyvemdmNotifiable
{
   // ...
}

class PluginFlyvemdmFleet extends PluginFlyvemdmNotifiable {
   // ...
}
```

The following code needs improvement 

https://github.com/flyve-mdm/glpi-plugin/blob/5006bceadab849a361d6c7987689153dd7fb3cce/inc/task.class.php#L415

the if / else () must be replaced by a call to a method defined by ```PluginFlyvemdmNotifiableInterface``` and implemented differently in ```PluginFlyvemdmFleet``` and ```PluginFlyvemdmAgent```

DIOHz0r:
### Changes description

New method created for notify an agent depending on the itemtype

### Checklist

### References

Closes #528

### DIFF
--- a/inc/fleet.class.php
+++ b/inc/fleet.class.php
@@ -368,12 +368,12 @@ class PluginFlyvemdmFleet extends CommonDBTM implements PluginFlyvemdmNotifiable
     * @return array instances of agents belonging to the fleet
     */
    public function getAgents() {
-      $id = $this->getID();
-      if (! ($id > 0)) {
+      if ($this->isNewItem()) {
          return [];
       }
       $agents = [];
       $agent = new PluginFlyvemdmAgent();
+      $id = $this->getID();
       $rows = $agent->find("`plugin_flyvemdm_fleets_id`='$id'");
 
       foreach ($rows as $row) {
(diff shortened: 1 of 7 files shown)


inc/fleet.class.php
```diff
@@ -555,4 +555,16 @@ public function isNotifiable() {
 
       return ($this->fields['is_default'] === '0');
    }
+
+   /**
+    * Find agents to be notified
+    * @param integer $notifiableId
+    * @return array
+    */
```
btry: Comment not needed. It should be in the interface file.

inc/notifiableinterface.class.php
```diff
@@ -87,4 +87,11 @@ public function refreshPersistedNotifications();
     * @return boolean True if message may be sent
     */
    public function isNotifiable();
+
+   /**
+    * Find agents to be notified
+    * @param integer $notifiableId
+    * @return array
+    */
+   public function findNotifiableAgents($notifiableId);
```
btry: GetNotifiableAgents() is a better name (because it tells the method returns something). This also contributes to similar method naming convention with Formcreator plugin (I introduced several methods having the same role for a similar need)

inc/fleet.class.php
```diff
@@ -368,12 +368,12 @@ public function cleanDBonPurge() {
     * @return array instances of agents belonging to the fleet
     */
    public function getAgents() {
-      $id = $this->getID();
-      if (! ($id > 0)) {
+      if ($this->isNewItem()) {
```
btry: extra refactor

inc/fleet.class.php
```diff
@@ -368,12 +368,12 @@ public function cleanDBonPurge() {
...
          return [];
       }
       $agents = [];
       $agent = new PluginFlyvemdmAgent();
+      $id = $this->getID();
```
btry: optimization

PR #555, opened 2018-06-18
New unit tests for Agents

DIOHz0r:
### Changes description

Added new tests for the agent class

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@diohz0r |99|

### DIFF
--- a/tests/suite-unit/PluginFlyvemdmAgent.php
+++ b/tests/suite-unit/PluginFlyvemdmAgent.php
@@ -45,6 +45,10 @@ class PluginFlyvemdmAgent extends CommonTestCase {
          case 'testShowDangerZone':
          case 'testPrepareInputForAdd':
          case 'testGetAgents':
+         case 'testAddDefaultJoin':
+         case 'testAddDefaultWhere':
+         case 'testGetTopic':
+         case 'testGetFleet':
             $this->login('glpi', 'glpi');
             break;
       }
@@ -60,6 +64,10 @@ class PluginFlyvemdmAgent extends CommonTestCase {
          case 'testShowDangerZone':
          case 'testPrepareInputForAdd':
          case 'testGetAgents':
+         case 'testAddDefaultJoin':
+         case 'testAddDefaultWhere':
+         case 'testGetTopic':
+         case 'testGetFleet':
             parent::afterTestMethod($method);
             \Session::destroy();
             break;
@@ -162,39 +170,43 @@ class PluginFlyvemdmAgent extends CommonTestCase {
     * @tags testShowForm
     * @engine inline
     */
-   /*public function testShowForm() {
+   public function testShowForm() {
+      $agent = $this->createAgent([]);
       $instance = $this->newTestedInstance();
       ob_start();
-      // TODO: have a fake agent registered in DB before this.
-      $instance->showForm(1);
+      $instance->showForm($agent->getID());
       $result = ob_get_contents();
       ob_end_clean();
+      $formAction = preg_quote("/plugins/flyvemdm/front/agent.form.php", '/');
       $this->string($result)
-         ->contains("method='post' action='-/plugins/flyvemdm/front/agent.form.php'")
+         ->matches("#method='post' action='.+?" . $formAction . "'#")
          ->contains("input type='hidden' name='entities_id' value='0'")
          ->contains("type='text' name='name'")
-         ->contains('input type="hidden" name="computers_id"')
-         ->contains('input type="hidden" name="plugin_flyvemdm_fleets_id"')
+         ->matches('#(?:input type="hidden"|select) .*?name="computers_id"#')
+         ->matches('#(?:input type="hidden"|select) .*?name="plugin_flyvemdm_fleets_id"#')
          ->contains('input type="hidden" name="_glpi_csrf_token"');
-   }*/
+   }
 
    /**
     * @tags testShowDangerZone
+    * @engine inline
     */
-   /*public function testShowDangerZone() {
+   public function testShowDangerZone() {
+      $agent = $this->createAgent([]);
       $instance = $this->newTestedInstance();
       ob_start();
-      // TODO: have a fake agent registered in DB before this.
-      $instance->showDangerZone(1);
+      $instance->showDangerZone($agent);
       $result = ob_get_contents();
       ob_end_clean();
+      $formAction = preg_quote("/plugins/flyvemdm/front/agent.form.php", '/');
       $this->string($result)
-         ->contains("method='post' action='/plugins/flyvemdm/front/agent.form.php'")
-         ->contains("input type='checkbox' class='new_checkbox' name='lock'")
-         ->contains("input type='checkbox' class='new_checkbox' name='wipe'")
+         ->matches("#method='post' action='.+?" . $formAction . "'#")
+         ->contains("input type='hidden' name='entities_id'")
+         ->matches("#input type='checkbox' .+? name='lock'#")
+         ->matches("#input type='checkbox' .+? name='wipe'#")
          ->contains('input type="submit" value="Unenroll" name="unenroll"')
          ->contains('input type="hidden" name="_glpi_csrf_token"');
-   }*/
+   }
 
    /**
     * @tags testShowForFleet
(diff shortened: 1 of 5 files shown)


tests/src/Glpi/Tests/CommonTestCase.php
```diff
@@ -145,10 +144,7 @@ protected function loginWithUserToken($userToken) {
     * Get a unique random string
     */
    protected function getUniqueString() {
-      if (is_null($this->str)) {
-         return $this->str = uniqid('str');
-      }
-      return $this->str .= 'x';
+      return uniqid('str', true);
```
btry: you should not change this implementation. It is picked from GLPI's unit tests and having the same implementation in both projects gives a change to mutualize the testing tools.

inc/agent.class.php
```diff
@@ -823,7 +823,7 @@ public function getSearchOptionsNew() {
          'table'              => $this->getTable(),
          'field'              => 'mdm_type',
          'name'               => __('MDM type', 'flyvemdm'),
-         'datatype'           => 'bool',
+         'datatype'           => 'specific',
```
btry: If you set ```specific``` here you must update getSpecificValueToDisplay(), and use getEnumMdmType() to map the value of the field into a localized equivalent.

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -161,39 +165,40 @@ public function testDefineTabs() {
...
       $instance = $this->newTestedInstance();
       ob_start();
-      // TODO: have a fake agent registered in DB before this.
-      $instance->showForm(1);
+      $instance->showForm($agent->getID());
       $result = ob_get_contents();
       ob_end_clean();
       $this->string($result)
-         ->contains("method='post' action='-/plugins/flyvemdm/front/agent.form.php'")
+         ->matches("#method='post' action='.+?\/plugins\/flyvemdm\/front\/agent\.form\.php'#")
```
btry: If you use a regex here, why not, but havilg a lot of backslashes makes harder to red it / debug it.

Consider using preg_quote http://php.net/manual/en/function.preg-quote.php to deletage the escaping and make long literal strings more readable, or an other method of your choice.

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -243,4 +255,96 @@ public function testGetAgents() {
          $this->boolean(isset($agents[$agent->getID()]))->isTrue();
       }
    }
+
+   /**
+    * @tags testAddDefaultJoin
+    */
+   public function testAddDefaultJoin() {
+      $instance = $this->newTestedInstance();
+      $result = $instance->addDefaultJoin();
```
btry: the method is static !

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -243,4 +255,96 @@ public function testGetAgents() {
...
+      $this->string($result)->isEmpty();
+
+   }
+
+   /**
+    * @tags testAddDefaultWhere
+    */
+   public function testAddDefaultWhere() {
+      $instance = $this->newTestedInstance();
+      $result = $instance->addDefaultWhere();
```
btry: the method is static !

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -243,4 +255,96 @@ public function testGetAgents() {
...
+
+   }
+
+   /**
+    * @tags testGetTopic
+    */
+   public function testGetTopic() {
+      $instance = $this->newTestedInstance();
+      $this->variable($instance->getTopic())->isNull();
+      // TODO: mock internal method to return the correct value
```
btry: no need to mock; just enroll an agent with the helper methods. As long you know the serial and the entity ID of your tested instance, you can determine  the expected topic.

When DI will be in place we can consider to mock safely. (by the way the related  PR very likely needs a rebase and an update)

Same comment applies to many test methods below

tests/suite-unit/PluginFlyvemdmInvitation.php
```diff
@@ -214,7 +214,7 @@ public function testPost_updateItem() {
       $instance = $this->newTestedInstance();
       $instance->add([
          '_useremails' => 'someone@localhost.local',
-         'entities_id' => 0,
+         'entities_id' => 10,
```
btry: ensure the entity exists first, to ensure DB consistency.

inc/agent.class.php
```diff
@@ -1982,6 +1982,10 @@ public static function getSpecificValueToDisplay($field, $values, array $options
                return $output;
             }
             break;
+         case 'mdm_type':
```
btry: This code will transform the field for API consumers (the dashboard). I think we should keep the data as is in such case, and delegate the dashboard to translate the data for display. isAPI() is your friend :+1:

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -35,6 +35,22 @@
...
+   /**
+    * @var \PluginFlyvemdmAgent
+    */
+   private $enrolledAgent = null;
+
+   /**
+    * Enrol an Agent for tests which need data
+    * @return \PluginFlyvemdmAgent
+    */
+   private function enrolTestAgent() {
```
btry: This method creates a potential dependency between tests. You should avoid it. Just create a new agent anytime you call this method.

Moreover when the tests of this file runs in parallell you have one process per test method and you cannot share the instance of the agent.

tests/suite-unit/PluginFlyvemdmInvitation.php
```diff
@@ -203,18 +203,26 @@ public function testShowForm() {
...
    }
 
+   /**
+    * @tags testPost_updateItem
+    */
    public function testPost_updateItem() {
+      $entity = new \Entity();
+      $entity->import([
+         'completename' => 'post update invitation ' . $this->getUniqueString(),
+      ]);
```
btry: perfect !

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -243,4 +277,102 @@ public function testGetAgents() {
...
+    * @tags testGetSpecificValueToDisplay
+    */
+   public function testGetSpecificValueToDisplay() {
+      $instance = $this->newTestedInstance();
+      $this->string($instance->getSpecificValueToDisplay('', ''))->isEqualTo('');
+      $this->string($instance->getSpecificValueToDisplay('is_online',
+         0))->contains('plugin-flyvemdm-offline');
+      $this->string($instance->getSpecificValueToDisplay('is_online',
+         1))->contains('plugin-flyvemdm-online');
+   }
```
btry: add a test for ```mdm_type``` field

tests/suite-unit/PluginFlyvemdmAgent.php
```diff
@@ -243,4 +277,102 @@ public function testGetAgents() {
...
+      $instance = $this->newTestedInstance();
+      $this->variable($instance->getComputer())->isNull();
+   }
+
+   /**
+    * @tags testRefreshPersistedNotifications
+    */
+   public function testRefreshPersistedNotifications() {
+      $instance = $this->newTestedInstance();
+      $this->variable($instance->refreshPersistedNotifications())->isNull();
```
btry: add a TODO comment to tell the test is incomplete. I guess this will be complex to test the other cases right now

tests/suite-unit/PluginFlyvemdmAgent.php
DIOHz0r: The idea of this method is for performance and only register one Agent on the DB for tests which need read a "real" data like showing a form or get mqtt topic. We change it to a mocked instance when DI is ready.

tests/suite-unit/PluginFlyvemdmAgent.php
DIOHz0r: done

tests/suite-unit/PluginFlyvemdmAgent.php
btry: > The idea of this method is for performance and only register one Agent on the DB

You don't have the expected optimization because the tests runs in parallel, in different processes. Then each test method (except one which has @engine inline annotation ) run in a distinct instance of the test class, each in a distinct process. 

please remove this method. Parallelization in atoum already speeds up things.

tests/suite-unit/PluginFlyvemdmEntityConfig.php
```diff
@@ -116,7 +116,7 @@ public function providerPrepareInputForUpdate() {
             [],
             [
                'You are not allowed to change the device limit',
-               'You are not allowed to download URL of the MDM agent',
+               'You are not allowed to change the download URL of the MDM agent',
```
btry: conflict to handle

btry:
Hi
This PR and #609 have some similar content. Is there still something interesting to pick here ?

DIOHz0r:
Hello, please review again this PR

tests/suite-unit/PluginFlyvemdmFusionInventory.php
```diff
@@ -35,6 +35,31 @@
...
+   }
+
+   /**
+    * @param $method
+    */
+   public function afterTestMethod($method) {
+      switch ($method) {
+         case 'testAddInvitationRule':
+         case 'testDeleteInvitationRuleCriteria':
+            parent::afterTestMethod($method);
```
btry: this should be moved before the switch, not in a case ```parent::afterTestMethod($method);```

btry:
Hi @DIOHz0r I added comments after the PR was merged.

I think the PR needs some more work. We may open an issue related to this PR to plan improvements I suggested.

PR #604, opened 2018-07-12
WIP Add locks to protect concurrent rules edition

btry:
Experimental - Add locks to protect concurrent rules edition

The purpose is to avoid simultaneous modification of rules in some specific cases. 

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@btry |2|

### References

Closes #N/A
Related #N/A
Depends on #N/A

### DIFF
--- a/inc/exception/entityruleeditionexception.class.php
+++ b/inc/exception/entityruleeditionexception.class.php
@@ -1,0 +1,6 @@
+<?php
+
+namespace GlpiPlugin\Flyvemdm\Exception;
+
+
+class EntityRuleEditionException extends \Exception {}
--- a/inc/fusioninventory.class.php
+++ b/inc/fusioninventory.class.php
@@ -29,7 +29,8 @@
  * ------------------------------------------------------------------------------
  */
 
- use GlpiPlugin\Flyvemdm\Exception\FusionInventoryRuleInconsistency;
+use GlpiPlugin\Flyvemdm\Exception\FusionInventoryRuleInconsistency;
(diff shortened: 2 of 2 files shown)


btry:
Hi @ingluife 

yes this is correct. If you read the PR to see the purpose is to ensure atomicity in some critical blocks of code (GLPI does not provides transactions yet, this is on its way)

DIOHz0r:
What do you mean with "modification of rules"? or in Which scenario this happens?

btry:
Consider there are 2 parallel executions to add an invitation. Due t o the lack of atomicity (transaction) for now, one process may create a rule to create computers in an entity while the other may do the same (with tthe same destination entity). 

In such scenario here is what might happen

* Process A searches for a rule to import a computer in entity 1 and gets 0 result 
* process B searches for a rule to import an a computer in the entity 1 and gets 0 result
* process A creates the rule because it does not exists yet
* Process B creates the rule because it does not exists yet

We finally have 2 rules doing the same (destination entity = 1) but we require to have only one.

Same type of scenario when removing criterias. If no criteria exists the rule is deleted. But this is risky if an other process tries to re-use the rule because it needs it.

We need locks to prevent this. The failure in unit tests you encouteered may be related to the lack of locks. Also I found il my local database duplicates rules making some tests fail. That's why I asked you if you also had duplicates rows (yesterday ?)

ingluife (dismissed):

btry:
FYI: https://github.com/glpi-project/glpi/commit/ba83cca20a5378378719fc78be1042f902071b65

Issue #601, opened 2018-07-11
Unit tests might fail in some condition


btry:
**Describe the bug**
An issue in unit tests:
```
> There is 1 error:
=> tests\units\PluginFlyvemdmPolicyDeployfile::testConflictCheck():
==> Error UNKNOWN in /home/travis/build/flyve-mdm/glpi/plugins/flyvemdm/tests/suite-unit/PluginFlyvemdmPolicyDeployfile.php on unknown line, generated by unknown file:
Fatal error: Uncaught Error: Call to a member function getID() on null in /home/travis/build/flyve-mdm/glpi/plugins/flyvemdm/tests/suite-unit/PluginFlyvemdmPolicyDeployfile.php:247
Stack trace:
#0 /home/travis/build/flyve-mdm/glpi/plugins/flyvemdm/vendor/atoum/atoum/classes/test.php(1363): tests\units\PluginFlyvemdmPolicyDeployfile->testConflictCheck()
#1 Standard input code(1): mageekguy\atoum\test->runTestMethod('testConflictChe...')
#2 {main}
  thrown in /home/travis/build/flyve-mdm/glpi/plugins/flyvemdm/tests/suite-unit/PluginFlyvemdmPolicyDeployfile.php on line 247
uncompleted
> There is 1 uncompleted method:
=> tests\units\PluginFlyvemdmPolicyDeployfile::testConflictCheck() with exit code 255:
==> output(153) "CommonDBTM::getFromDBByCrit() in /home/travis/build/flyve-mdm/glpi/inc/commondbtm.class.php line 251
==> getFromDBByCrit expects to get one result, 2 found!"
> Total tests duration: 5.26 seconds.
> Total tests memory usage: 91.23 Mb.
> Running duration: 17.66 seconds.
```
**To Reproduce**
There is maybe a race condition when unit tests run in parallel

DIOHz0r:
I'm also getting this error for other tests...

```
  *** MySQL query error:
  SQL: INSERT INTO `glpi_plugin_fusioninventory_dblocksoftwares`
               SET `value`='1'
  Error: Duplicate entry '1' for key 'PRIMARY'
  Backtrace :
  ...ventory/inc/inventorycomputerlib.class.php:2509 
  ...nventory/inc/inventorycomputerlib.class.php:822 PluginFusioninventoryInventoryComputerLib->importSoftware()
[...]

PR #609, opened 2018-07-13
Refactor code for keep or change session on tests

DIOHz0r:
### Changes description

I found a bug for testing which was mixing the session of glpi user and guest user for enrol agents. this refactor fix that.

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@diohz0r |8|

### References

Related #601

### DIFF
--- a/tests/before_script.sh
+++ b/tests/before_script.sh
@@ -11,14 +11,6 @@ if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then
   composer config -g github-oauth.github.com $GH_OAUTH
 fi
 
-# Make a ramdisk for mysql (speed improvement)
-sudo mkdir /mnt/ramdisk
-sudo mount -t tmpfs -o size=1024m tmpfs /mnt/ramdisk
-sudo stop mysql
-sudo mv /var/lib/mysql /mnt/ramdisk
-sudo ln -s /mnt/ramdisk/mysql /var/lib/mysql
-sudo start mysql
-
 # setup GLPI and its plugins
 mysql -u root -e 'create database $DBNAME;'
 git clone --depth=35 $GLPI_SOURCE -b $GLPI_BRANCH ../glpi && cd ../glpi
--- a/tests/src/Flyvemdm/Tests/CommonTestCase.php
+++ b/tests/src/Flyvemdm/Tests/CommonTestCase.php
@@ -28,6 +28,7 @@
  * @link      https://flyve-mdm.com/
  * ------------------------------------------------------------------------------
  */
+
 namespace Flyvemdm\Tests;
 
 use Glpi\Tests\CommonTestCase as GlpiCommonTestCase;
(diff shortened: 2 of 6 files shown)


tests/src/Flyvemdm/Tests/CommonTestCase.php
```diff
@@ -39,25 +40,38 @@ class CommonTestCase extends GlpiCommonTestCase {
...
+         // item has been created but its data is incomplete so let's load it
+         $agent->getFromDB($agent->getID());
+      }
+
+      if ($keepLoggedUser && $currentUser) {
+         // Go back to previous logged user
+         $this->terminateSession();
+         $this->restartSession();
+         $this->setupGLPIFramework();
+         $this->login($currentUser, $currentUser);
```
btry: this will not work with all user accounts.

tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -92,144 +96,118 @@ public function testDeviceCountLimit() {
...
-      $agent = $this->agentFromInvitation($user, $guestEmail, $serial, 'bad token');
-      $this->boolean($agent->isNewItem())
-         ->isTrue(json_encode($_SESSION['MESSAGE_AFTER_REDIRECT'], JSON_PRETTY_PRINT));
-      $this->array($_SESSION['MESSAGE_AFTER_REDIRECT'][ERROR])->contains('Invitation token invalid');
-
-      // Test the invitation log did not increased
-      // this happens because the enrollment failed without identifying the invitation
-      $invitationLog = new \PluginFlyvemdmInvitationlog();
-      $rows = $invitationLog->find("`plugin_flyvemdm_invitations_id` = '$inviationId'");
-      $this->integer(count($rows))->isEqualTo($expectedLogCount);
```
btry: We loose the tests on invitation logs. Can you re-add it while keeping the factorization ?

tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -92,144 +96,118 @@ public function testDeviceCountLimit() {
...
       // Test successful enrollment
       $agent = $this->agentFromInvitation($user, $guestEmail, $serial, $invitationToken, 'apple');
       $this->boolean($agent->isNewItem())
          ->isFalse(json_encode($_SESSION['MESSAGE_AFTER_REDIRECT'], JSON_PRETTY_PRINT));
 
-      // Test there is no new entry in the invitation log
-      $invitationLog = new \PluginFlyvemdmInvitationlog();
-      $fk = \PluginFlyvemdmInvitation::getForeignKeyField();
-      $rows = $invitationLog->find("`$fk` = '$inviationId'");
-      $this->integer(count($rows))->isEqualTo($expectedLogCount);
```
btry: we need to keep check in invitation log

tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -937,8 +902,10 @@ private function loginAndAddAgent(array $currentInvitation) {
          'version'           => \PluginFlyvemdmAgent::MINIMUM_ANDROID_VERSION . '.0',
          'type'              => 'android',
          'inventory'         => CommonTestCase::AgentXmlInventory($serial),
-      ]);
-      return $agentId;
+      ];
+      $agent = $this->enrollFromInvitation($userId, $input);
+
+      return (int)$agent->getID();
    }
```
btry: I think we can use a helper method : this one creates an agent as well.

tests/src/Flyvemdm/Tests/CommonTestCase.php
DIOHz0r: I know that but we can't now the password and also the test are using glpi user for login right now, We can set the same username and password for our test.

btry: someday we may need other use cases. We may add a test to check the login succeeded. This way we immediately know why subsequent execution flow fails, if it happens.

tests/suite-integration/PluginFlyvemdmAgent.php
DIOHz0r: I updated this but the logic has changed a little bit

tests/src/Flyvemdm/Tests/CommonTestCase.php
DIOHz0r: I added one asserter for that case

tests/suite-integration/PluginFlyvemdmAgent.php
btry: No problem  !

tests/before_script.sh
```diff
@@ -11,14 +11,6 @@ if [ "$TRAVIS_SECURE_ENV_VARS" = "true" ]; then
...
 fi
 
-# Make a ramdisk for mysql (speed improvement)
-sudo mkdir /mnt/ramdisk
-sudo mount -t tmpfs -o size=1024m tmpfs /mnt/ramdisk
-sudo stop mysql
-sudo mv /var/lib/mysql /mnt/ramdisk
-sudo ln -s /mnt/ramdisk/mysql /var/lib/mysql
-sudo start mysql
-
```
btry: Do you observe a change in job execution duration ?

DIOHz0r: Errors were more consistent and running times aren't too much different

btry: If time don't change a lot, OK to drop. However this shall not affect tests. OK to drop

PR #614, opened 2018-07-16
 Fix message of FusionInventory agent not created

### LINKED ISSUE #610: FusionInventory agent not created
Hi, @btry 
No inventory in the Web Dashboard.
No computer in the GLPI.
Agent said: Computer P2.
Error in the enrollment: FusionInventory agent not created

Inventory exported as XML and JSON.
[inventory-p2.zip](https://github.com/flyve-mdm/android-mdm-agent/files/2194494/inventory-p2.zip)

We need to submit a path to FI
Or debug in deep what happen.

References:

image

image

Related to:
https://github.com/flyve-mdm/android-mdm-agent/issues/504

DIOHz0r:
### Changes description

Before the fix the plugin tried to get the FI agent using the session var 'plugin_fusioninventory_agents_id' which is not the real id registered on DB. Now it looks for it using the UUID on the inventory which matches with the agent name 

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHz0r |21|
|@Rafaelje |6|

### References

Closes #610


inc/agent.class.php
```diff
@@ -440,7 +440,8 @@ public function prepareInputForAdd($input) {
       }
 
       // User already logged in : user token has been validated
-
+      ini_set("memory_limit", "-1");
+      ini_set("max_execution_time", "0");
```
btry: Is this really needed ? If I remember correctly, Fusion already does it

btry: ok, then let's keep it

PR #618, opened 2018-07-18
Fix for plugin upgrade

### LINKED ISSUE #616: Class not found on upgrade
**Describe the bug**
A problem loading a MQTT class is throwing when the plugin need an update

**To Reproduce**
1. Go to php command line and execute ```php tools\cli_install.php```
2. See error

**Expected behavior**
An upgrade for the plugin

**Screenshots**
imagen

DIOHz0r:
### Changes description

Update the plugin setup script which missed to include the autoloader on some status

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHz0r |4|

### References

Closes #616

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -90,8 +90,17 @@ function plugin_init_flyvemdm() {
       ]);
    }
 
-   if ($plugin->isActivated('flyvemdm')) {
+   if (!$plugin->getFromDBbyDir('flyvemdm')) {
+      // nothing more to do at this moment
+      return;
+   }
+
+   $state = $plugin->getField('state');
+   if ($state != Plugin::NOTACTIVATED) {
       require_once(__DIR__ . '/vendor/autoload.php');
+   }
+
+   if ($state == Plugin::ACTIVATED) {
       if (!class_exists('GlpiLocalesExtension')) {
          require_once(__DIR__ . '/lib/GlpiLocalesExtension.php');
       }


setup.php
```diff
@@ -90,8 +90,17 @@ function plugin_init_flyvemdm() {
       ]);
    }
 
-   if ($plugin->isActivated('flyvemdm')) {
+   if(!$plugin->getFromDBbyDir('flyvemdm')){
+      // nothing more to do at this moment
+      return;
```
btry: I think you don't need to check if the method succeeded. GLPI automatically adds an entry in the DB if the plugin exists in the filesystem. The only case where a return will occur is when GLPI fails to write the row (mysql database with full filesystem or other severe failure not related to GLPI or its plugins)

inc/mqttclient.class.php
```diff
@@ -282,7 +282,7 @@ protected function getMQTTConnection() {
    protected function buildMqtt($socketAddress, $port, $isTls, $sslCipher) {
       $protocol = $isTls ? "ssl://" : "tcp://";
       try {
-         $mqtt = new sskaje\mqtt\MQTT("$protocol$socketAddress:$port");
+         $mqtt = new \sskaje\mqtt\MQTT("$protocol$socketAddress:$port");
```
btry: you can remove this change.

setup.php
DIOHz0r: There are two checks, the first one is for not execute the rest of the checks because if that query return false and I don't check it, the autoload load will be include by the reason of the field state will return N/A which will be different of NOTACTIVATED

PR #652, opened 2018-07-21
Fix upgrade

btry:
### Changes description

Introduce a schema version of the database for the plugin, which should match the 2 major numbers of the version of the plugin. The schema shall not change for bugfixes.

Also, the release rc.1 still has the update_to_dev migration script, this was a miss. 

TODO:
add a check in Robo file to detect this file, or each time we reelase, bump version for the development branch and make upgrade_to_xx script . To be refined...

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@btry |3|

### References

Closes #N/A
Related #N/A
Depends on #N/A

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -33,12 +33,15 @@
  * Entry point for installation process
  */
 function plugin_flyvemdm_install() {
-   global $DB;
-
+   $version   = plugin_version_flyvemdm();
+   $migration = new Migration($version['version']);
    require_once(PLUGIN_FLYVEMDM_ROOT . "/install/installer.class.php");
+   spl_autoload_register([__CLASS__, 'autoload']);
    $installer = new PluginFlyvemdmInstaller();
-
-   return $installer->install();
+   if (!$install->isPluginInstalled()) {
+      return $installer->install($migration);
+   }
+   return $installer->upgrade($migration);
 }
 
 /**
--- a/hook.php
+++ b/hook.php
@@ -35,13 +35,13 @@
 function plugin_flyvemdm_install() {
    $version   = plugin_version_flyvemdm();
    $migration = new Migration($version['version']);
-   require_once(PLUGIN_FLYVEMDM_ROOT . "/install/installer.class.php");
-   spl_autoload_register([__CLASS__, 'autoload']);
-   $installer = new PluginFlyvemdmInstaller();
+   require_once(PLUGIN_FLYVEMDM_ROOT . "/install/install.class.php");
+   spl_autoload_register([PluginFlyvemdmInstall::class, 'autoload']);
+   $install = new PluginFlyvemdmInstall();
    if (!$install->isPluginInstalled()) {
-      return $installer->install($migration);
+      return $install->install($migration);
    }
-   return $installer->upgrade($migration);
+   return $install->upgrade($migration);
 }
 
 /**
@@ -49,10 +49,10 @@ function plugin_flyvemdm_install() {
  * @return boolean True if success
  */
 function plugin_flyvemdm_uninstall() {
-   require_once(PLUGIN_FLYVEMDM_ROOT . "/install/installer.class.php");
-   $installer = new PluginFlyvemdmInstaller();
+   require_once(PLUGIN_FLYVEMDM_ROOT . "/install/install.class.php");
+   $install = new PluginFlyvemdmInstall();
 
-   return $installer->uninstall();
+   return $install->uninstall();
 }
 
 /**
--- a/install/install.class.php
+++ b/install/install.class.php
@@ -39,7 +39,7 @@ if (!defined('GLPI_ROOT')) {
  * @since 0.1.0
  *
  */
-class PluginFlyvemdmInstaller {
+class PluginFlyvemdmInstall {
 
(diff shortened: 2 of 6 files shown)


hook.php
```diff
@@ -33,12 +33,14 @@
  * Entry point for installation process
  */
 function plugin_flyvemdm_install() {
-   global $DB;
-
```
btry: $DB not used

hook.php
```diff
@@ -33,12 +33,14 @@
...
+   $migration = new Migration($version['version']);
    require_once(PLUGIN_FLYVEMDM_ROOT . "/install/installer.class.php");
    $installer = new PluginFlyvemdmInstaller();
-
-   return $installer->install();
+   if (PluginFlyvemdmInstaller::getCurrentVersion() == '') {
+      return $installer->install($migration);
+   } 
+   return $install->upgrade($migration);
 }
```
btry: do not delegate to the method install() the task to distinguish a fresh installation and an upgrade

install/installer.class.php
```diff
@@ -45,30 +45,6 @@ class PluginFlyvemdmInstaller {
...
-      'notificationtargetinvitation',
-      'geolocation',
-      'policy',
-      'policycategory',
-      'fleet_policy',
-      'wellknownpath',
-      'invitation',
-      'invitationlog',
-   ];
-
```
btry: this array is no longer used

install/installer.class.php
```diff
@@ -136,7 +101,12 @@ public function install() {
       $this->createRootEntityConfig();
       $this->createDisplayPreferences();
 
-      Config::setConfigurationValues('flyvemdm', ['version' => PLUGIN_FLYVEMDM_VERSION]);
+      Config::setConfigurationValues(
+         'flyvemdm', [
+            'version' => PLUGIN_FLYVEMDM_VERSION,
+            'schema_version' => PLUGIN_FLYVEMDM_SCHEMA_VERSION,
+         ]
+      );
```
btry: at the end of an installation we must update the version and schema version of the plugin

install/installer.class.php
```diff
@@ -494,20 +464,74 @@ public function createNotificationTargetInvitation() {
...
+      $this->migration = $migration;
+      $fromSchemaVersion = $this->getSchemaVersion();
+
+      switch ($fromSchemaVersion) {
+         case '2.0':
+            // Example : upgrade to schema version 3.0
+            // $this->upgradeOneStep('3.0');
+         case '3.0':
+            // Example : upgrade to version 4.0
+            // $this->upgradeOneStep('4.0');
```
btry: versions used here are now schema versions

install/installer.class.php
```diff
@@ -494,20 +464,74 @@ public function createNotificationTargetInvitation() {
...
+      $this->createRootEntityConfig();
+      $this->createDisplayPreferences();
+
+      Config::setConfigurationValues(
+         'flyvemdm', [
+            'version' => PLUGIN_FLYVEMDM_VERSION,
+            'schema_version' => PLUGIN_FLYVEMDM_SCHEMA_VERSION,
+         ]
+      );
+   }
```
btry: here too: we must update version and schema version

install/installer.class.php
```diff
@@ -494,20 +464,74 @@ public function createNotificationTargetInvitation() {
...
+
+   /**
+    * Find the version of the plugin
+    *
+    * @return string|NULL
+    */
+   protected function getSchemaVersion() {
+      return $this->getSchemaVersionFromGlpiConfig();
+   }
```
btry: this method might be useful to handle schema version detection without relying on the data in the configuration (when we find a bug lately).

install/installer.class.php
```diff
@@ -494,20 +464,66 @@ public function createNotificationTargetInvitation() {
...
+            // $this->upgradeOneStep('3.0');
+         case '3.0':
+            // Example : upgrade to version 4.0
+            // $this->upgradeOneStep('4.0');
 
          default:
       }
-      if (PluginFlyvemdmCommon::endsWith(PLUGIN_FLYVEMDM_VERSION, "-dev")) {
-         $this->upgradeOneStep('dev');
+
```
btry: (strange: the closing bracket is missing in the diff!)

I decided to obsolete upgrade_to_dev.php to avoid upgrade misses. Now we rely only on the schema version for upgrades.

install/install.class.php
```diff
@@ -515,36 +474,54 @@ public function createNotificationTargetInvitation() {
...
             // Unknown version
+            return false;
       }
-      if (is_readable(__DIR__ . '/upgrade_to_develop.php')
-         && PluginFlyvemdmCommon::endsWith(PLUGIN_FLYVEMDM_VERSION, 'develop')
-      ) {
-         $upgradeStep = new PluginFlyvemdmUpgradeToDevelop();
-         $upgradeStep->upgrade(new Migration('develop'));
-      }
+      $this->upgradeOneStep('develop');
```
DIOHz0r: This will make the plugin upgrade always to develop, won't it?

btry: Yes, but only if the upgrade_to_develop.php file exists. If it does not exists (this is true in a release) then no problem.

PR #683, opened 2018-08-08
WIP Set development version

btry:
the purpose is to have a fixed version name for develop branch, to prepare a better use of git flow

This version is not semver compliant, but I think this is not a problem as develop is never used as is for a release.

When we want to prepare a release we can create the release/ branch based on develop, stabilize it and produce release candidates. The release/ branch will bump version at lease once, and we cannot bump the version in develop branche without having a conflict when finishing a release/ branch. Using a invariable version like **develop** we prevent any merge conflict, advertise this version is not for production use.

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@btry |2|

### References

Closes #N/A
Related #N/A
Depends on #N/A


btry:
Build failed because *develop* is not semver and composer reads it when running in the project's directory, even for commands not related to the project (self-upgrade in Travis)

Checking the documentation of packagist and npm I decided to remove the version property:
https://getcomposer.org/doc/04-schema.md#version
https://docs.npmjs.com/files/package.json
- composer:  ```Packagist uses VCS repositories, so the statement above is very much true for Packagist as well. Specifying the version yourself will most likely end up creating problems at some point due to human error.```
- npm: ```If you don't plan to publish your package, the name and version fields are optional.```

indeed, the project will not be published on npmjs, because this is not a nodeJS project.

This will make build simpler in Robo, and workaround the issue I encouteer right now.

package.json
```diff
@@ -2,7 +2,6 @@
     "name": "@teclib/flyve-mdm-glpi",
     "version": "2.0.0-rc.1",
     "description": "Flyve MDM plugin for GLPI",
-    "main": "config.js",
```
btry: this property should notexists, because this is not a node package !

btry: main is relevant for node packages only : this is an entry point

PR #690, opened 2018-08-10
Task and Taskstatus data for tests

DIOHz0r:

### Changes description

This PR allows the creation of data needed for some tests. 

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHz0r |9|

### References

Related #N/A

### DIFF
--- a/tests/src/Flyvemdm/Tests/CommonTestCase.php
+++ b/tests/src/Flyvemdm/Tests/CommonTestCase.php
@@ -187,6 +187,56 @@ class CommonTestCase extends GlpiCommonTestCase {
    }
 
    /**
+    * @param array $input input data
+    * @return \PluginFlyvemdmTask
+    */
+   protected function createTask($input) {
+      $task = $this->newMockInstance(\PluginFlyvemdmTask::class, '\MyMock');
+      $task->getMockController()->post_addItem = function () {};
+      $taskId = $task->add($input);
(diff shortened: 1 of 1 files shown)


tests/src/Flyvemdm/Tests/CommonTestCase.php
```diff
@@ -186,6 +186,56 @@ protected function createFleet($input) {
...
+
+      return $task;
+   }
+
+   /**
+    * @param string $policySymbol
+    * @return array
+    */
+   protected function createFleetAndTask($policySymbol = 'storageEncryption') {
+      $fleet = $this->createFleet(['name' => $this->getUniqueString()]);
```
btry: Remove this method, the test should call createFleet, then createTask.

DIOHz0r: This method is a from a code refactor used on createAgentTaskstatus and two more test that haven't been published yet because the code is on another PR.

btry: When a method has a name telling that ot dies several things, it is better to split it.

The caller could do the following:
```php
$fleet = $this->createFleet();
$task = $this->createTask($fleet);
// test whatever with these 2 above objects
```
```

btry: Ok, I approve, and let's see the final result !

PR #734, opened 2018-09-05
Add CI script for documentation

DIOHz0r:
### Changes description

I've noticed that after generated RC2 the documentation for master wasn't updated because the script change was not on the release branch. This must fix the issue

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHz0r |2|

### References

Closes #N/A
Related #N/A
Depends on #N/A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
     - GLPI_SOURCE="https://github.com/glpi-project/glpi"
     - FI_SOURCE="https://github.com/fusioninventory/fusioninventory-for-glpi"
     - CS=7.2
+    - APIDOCS_PHP=7.1
     - DBNAME=glpitest
     - OLDDBNAME=glpiupgradetest
     - AFTER_SUCCESS_BRANCH=9.3/bugfixes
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -57,6 +57,7 @@ script:
 after_success:
   # let's update the documentation and locales
   - if [ ${TRAVIS_PHP_VERSION:0:3} == "$CS" ] && [ "$GLPI_BRANCH" = "$AFTER_SUCCESS_BRANCH" ]; then tests/after_success.sh; fi
+  - if [ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ] && [ "$GLPI_BRANCH" = "$AFTER_SUCCESS_BRANCH" ]; then tests/apidocs.sh; fi
```
btry: prefer a variable like we do with $CS, this is more maintainable

DIOHz0r: This is the only place used for check the PHP 7.1 version, is really necessary?

btry: this is for convenience, when php will raise its version. This also gives the opportunity to comment the variable and tell it must be different from $CS to ensure apigen does not conflicts with other dependencies in composer.json

btry:
@DIOHz0r 

I think we can / should cherry-pick this pr into develop. Waiting for your approval.

PR #745, opened 2018-09-12
Add script for systemd in wizard

Naylin15:
### Changes description
Add path to script for systemd

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@Naylin15 |1|

### References

Closes #N/A
Related #N/A
Depends on #N/A


tpl/config-wizard.html.twig
```diff
@@ -245,6 +245,7 @@ ciphers ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES2
 				<p>{{ __('Check the user configured in the script and adapt it if needed.', 'flyvemdm') }}</p>
 				<p>update-rc.d flyvemdm defaults</p>
 				<p>service flyvemdm start</p>
+				<p>{{ __('For systemd, use the script located at /var/www/html/glpi/plugins/flyvemdm/scripts/flyvemdm.service', 'flyvemdm') }}</p>
```
btry: we should not give a canonical path because it may vary. Also we must tell the user, group and path must be changed.

I think we can improve the SysV and systemD files by doing like we do with config.js for the dashboard: naming them script.example.sh and flyve.example.service

PR #753, opened 2018-09-18
Add MQTT ports for backend

### LINKED ISSUE #751: Allow different MQTT ports for devices and plugin
We already allow distinct host for mqtt, but we enforce the same port. let's extend the feature to ports also.

Naylin15:
### Changes description
Add MQTT ports for backend

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@Naylin15 |4|

### References

Closes #751 
Related #N/A
Depends on #N/A


install/install.class.php
```diff
@@ -607,6 +607,8 @@ protected function createInitialConfig() {
          'mqtt_broker_internal_address'    => '127.0.0.1',
          'mqtt_broker_port'                => '1883',
          'mqtt_broker_tls_port'            => '8883',
+         'mqtt_broker_port_backend'        => '1884',
+         'mqtt_broker_tls_port_backend'    => '8884',
```
btry: I suggest the defaults for t hose 2 settings be the same as the existing ones : 1883 and 8883.

For most configurations they will be the same. Different values may be required for some very specific cases.

tpl/config-messagequeue.html.twig
```diff
@@ -27,13 +21,34 @@
...
+         <td>{{ __('MQTT broker address', 'flyvemdm') }}</td>
+         <td><input type="text" name="mqtt_broker_internal_address"
+            value="{{ config.mqtt_broker_internal_address }}"></td>
+         <td>{{ __('An IP address or a hostname used by the backend to communicate with the broker', 'flyvemdm') }}</td>
+      </tr>
+      <tr class="tab_bg_1">
+         <td>{{ __('MQTT broker port', 'flyvemdm') }}</td>
+         <td><input type="number" name="mqtt_broker_port_backend" value="{{ config.mqtt_broker_port_backend }}"
+            min="1" max="65535"></td>
+         <td>{{ __('A port number between 1025 and 65535, standard port is 1884', 'flyvemdm') }}</td>
```
btry: same here. By the way the word standard refers to IANA:  https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=mqtt

btry:
Hi @Naylin15 

just in case, the upgrade code needs to pick populate the ports for the backend by copying the ports for devices.

Naylin15:
> Hi @Naylin15
> 
> just in case, the upgrade code needs to pick populate the ports for the backend by copying the ports for devices.

Hi @btry 

Fixed the port numbers but I did not understand this comment, the part of copying them for the devices

Do you mean to get the port of the devices (clients) and set them for backend?

btry:
> I did not understand this comment, the part of copying them for the devices

> Do you mean to get the port of the devices (clients) and set them for backend?

Yes. add this required code in the upgrade from 2.0 to 2.1.

Naylin15:
Hi @btry

I think this PR is ready

Thank you

btry:
The feature looks good. We need to test i in a functional way
- fresh install
- 2.0 branch upgraded to develop

I'm confident this will work; I may test it on an instance I use.

PR #757, opened 2018-09-21
MQTTlog update

DIOHz0r:

### Changes description

This update is for the changes made on the last PR about mqtt features.

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHZ0r |31|

### References

Related #754


btry:
@ajsb85 don't approve now: if we have consensus about my proposal, the dev is not finished ;)

DIOHz0r:
> truncate the beginning of a topic with the topic of the notifiable ($notifiable->getTopic()) before saving it to the table (mqtthandler class)

The function where the log is saved the notifiable object is not available for use because its isolated logic, therefore we will need to add more logic to get the real itemtype and the item_id for outgoin and incoming mqtt messages

btry:
There is an interface NotifiableInterface. I dont have access to source right now. We can  use (or create) an abstract class Notifiable with a static method getItemtypeByTopic() to find the missing information. Next you can instantiate from its result and call getByTopic() which exists in the notifiable interface

tests/src/Glpi/Tests/CommonDBTestCase.php
```diff
@@ -188,7 +188,8 @@ public function checkInstall($filename = '', $filter = 'glpi_', $when = '') {
                $s_type[0] = str_replace(" COLLATE utf8_unicode_ci", "", $s_type[0]);
                $s_type[0] = str_replace(" CHARACTER SET utf8", "", $s_type[0]);
                if (strpos(trim($s_type[0]), 'text') === 0
-                   || strpos(trim($s_type[0]), 'longtext') === 0) {
+                  || strpos(trim($s_type[0]), 'mediumtext') === 0
```
btry: This method is copied from GLPI. It is not available in GLPI 9.2 then I planned to remive it when we will support GLPI 9.3 minimum.

Then we need to apply this fix to GLPI also.

tests/src/Glpi/Tests/CommonDBTestCase.php
```diff
@@ -188,7 +188,8 @@ public function checkInstall($filename = '', $filter = 'glpi_', $when = '') {
...
+                  || strpos(trim($s_type[0]), 'longtext') === 0) {
```
btry: i thought I disabled the call to this method long ago.

GLPI provides tools to compare an upgraded DB against a fresh install DB. This is more useful than this stuf, which only checks if the empty sql file was correctly processed.

Please re-move this method and its call. This is obsolete.

inc/mqttlog.class.php
```diff
@@ -100,10 +100,36 @@ protected function saveMqttMessage($direction, $topicList, $msg) {
       $msg = $DB->escape($msg);
       foreach ($topicList as $topic) {
          $topic = $DB->escape($topic);
+
+         $chunks = explode('/', $topic, 4);
+         if ($chunks[0] == '' || !isset($chunks[3])) {
+            // avoid save invalid topic strings
```
btry: I think you should describe more explictly the messages you want to filter.

DIOHz0r: done

inc/mqttlog.class.php
```diff
@@ -100,10 +100,37 @@ protected function saveMqttMessage($direction, $topicList, $msg) {
       $msg = $DB->escape($msg);
       foreach ($topicList as $topic) {
          $topic = $DB->escape($topic);
+
+         $chunks = explode('/',$topic,4);
```
btry: missing space after comma

install/upgrade_to_2_0.php
```diff
@@ -355,6 +355,32 @@ public function upgrade(Migration $migration) {
...
       $migration->dropTable($table);
 
+      $table = 'glpi_plugin_flyvemdm_mqttlogs';
+      $migration->changeField($table, 'message', 'message', 'mediumtext');
+      $migration->addField($table, 'itemtype', 'string', ['after' => 'message']);
+      $migration->addField($table, 'items_id', 'integer', ['after' => 'itemtype']);
+      $migration->addKey($table, 'itemtype', 'itemtype');
+      $migration->addKey($table, 'items_id', 'items_id');
+      if (!$DB->fieldExists($table, 'name')) {
+         // fix fleets logs
```
btry: this is not a fix.. this is not a feature, this is .. an upgrade !

install/upgrade_to_2_0.php
```diff
@@ -355,6 +355,32 @@ public function upgrade(Migration $migration) {
...
+         $migration->addPostQuery("UPDATE $table as t1,
+            (SELECT id, SUBSTRING_INDEX(SUBSTRING_INDEX(topic, '/', 3), '/', -1) as new_items_id,
+              SUBSTRING(REPLACE(topic, SUBSTRING_INDEX(topic, '/', 3), ''), 2) as new_topic
+              FROM $table WHERE topic NOT LIKE '/%' and topic like '%/fleet/%') as t2
+            SET t1.itemtype = 'PluginFlyvemdmFleet', t1.items_id = t2.new_items_id, 
+            t1.topic = t2.new_topic, t1.topic = t2.new_topic WHERE t1.id = t2.id");
+
+         // fix agents logs
```
btry: same here :)

tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -612,7 +612,7 @@ public function testPingRequest() {
       $logEntryFound = 0;
       $rows = $log->find("`direction` = 'O' AND `id` > '$lastLogId'");
       foreach ($rows as $row) {
-         if ($row['topic'] == $agent->getTopic() . '/Command/Ping') {
+         if ($row['topic'] == 'Command/Ping') {
```
btry: without this PR the tests select the log entry for the tested agent. Now, the topic does not contains this information. I think the unit  test is degraded by your change. We should add some assertions to ensure we will not wrongly pick a message for an other item.

tests/suite-integration/PluginFlyvemdmAgent.php
```diff
@@ -700,7 +700,7 @@ public function testInventoryRequest() {
       $logEntryFound = 0;
       $rows = $log->find("`direction` = 'O' AND `id` > '$lastLogId'");
       foreach ($rows as $row) {
-         if ($row['topic'] == $agent->getTopic() . '/Command/Inventory') {
+         if ($row['topic'] == 'Command/Inventory') {
```
btry: without this PR the tests select the log entry for the tested agent. Now, the topic does not contains this information. I think the unit  test is degraded by your change. We should add some assertions to ensure we will not wrongly pick a message for an other item.

Same comment for several other changes below in unit tests.

install/upgrade_to_2_0.php
DIOHz0r: I don't get the meaning for this comment and the next one, there is no going to be an action for migrate the data?

btry: this comment is about the "fix fleets logs" sentence. This code does not fixes a bug, it updates the content of a table to match a new specification for logs.

I think "fix" is not accurate here

tests/plugin_flyvemdm_empty_2.0.0-rc.1.sql
```diff
@@ -104,7 +104,9 @@ CREATE TABLE IF NOT EXISTS `glpi_plugin_flyvemdm_mqttlogs` (
   `date` datetime DEFAULT NULL,
   `direction` varchar(1) NOT NULL DEFAULT '' COMMENT 'I for received message, O for sent message',
   `topic` varchar(255) NOT NULL DEFAULT '',
-  `message` varchar(255) DEFAULT NULL,
+  `message` mediumtext DEFAULT NULL,
+  `itemtype` varchar(255) DEFAULT NULL,
+  `items_id` int(11) NOT NULL DEFAULT '0',
   PRIMARY KEY (`id`),
   KEY `date` (`date`),
   KEY `topic` (`topic`)
```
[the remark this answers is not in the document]
DIOHz0r: thank you I though that I did that.

PR #758, opened 2018-09-21
Disable MQTT messages on wrong app filename

### LINKED ISSUE #748: Application name is null
**Describe the bug**
When adding a new application on Dev environment the tag deployApp on the MQTT message come with a null value

**To Reproduce**
1. Go to Flyve
2. Click on General -> Package
3. Add new application
4. Create a fleet and add the applications
5. Add the fleet to the device
6. Open the Android Agent Activity -> Log

**Expected behavior**
Get the deployApp with the package name of the app or the name of the app

**Screenshots**
screenshot_1536962255

DIOHz0r:

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHz0r |22|

### References

Closes #748

### DIFF
--- a/inc/package.class.php
+++ b/inc/package.class.php
@@ -227,7 +227,7 @@ class PluginFlyvemdmPackage extends PluginFlyvemdmDeployable {
       // Find the added file
       $preparedFile = $this->prepareFileUpload();
 
-      if (!$preparedFile || !is_array($preparedFile)) {
+      if (!$preparedFile || !is_array($preparedFile) || !array_filter($preparedFile)) {
          return $input;
       }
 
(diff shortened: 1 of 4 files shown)


DIOHz0r:
> Not sending the MQTT message require we send it later. I enumerated some cases.

The message will not be sent by the parser after the update of the package info? That's why the only notification call is in the post update action.

btry:
> The message will not be sent by the parser after the update of the package info?

No, the parser does not triger a mqtt message itself. However the parser calls $package->update() then post_updateItem will run. The mesage should be sent.

Need a real world test to validate, and if OK, we can merge.

DIOHz0r:
> the parser does not triger a mqtt message itself

With these changes now it does when the metadata of the apk is updated :)

btry:
If you add unit tests related to this feature, they should be in this PR. Any other unit test must go in an other PR.

inc/package.class.php
```diff
@@ -471,12 +481,17 @@ private function parseApplication() {
       $manifest = $apk->getManifest();
       $iconResources = $apk->getResources($manifest->getApplication()->getIcon());
       $apkLabel = $apk->getResources($manifest->getApplication()->getLabel());
-      $input['icon'] = base64_encode(stream_get_contents($apk->getStream($iconResources[0])));
+      $stream = $apk->getStream($iconResources[0]);
+      if (!is_resource($stream)) {
+         return false;
```
btry: This is quite aggressive to fail when we cannot extract the icon. Instead I suggest to put a blank image. One transparent pixel with PNG should be OK. Hard codable in a constant and small.

btry:
Merge also in develop please. 

git  checkout develop
git pull origin develop
git  merge --no-ff feature/my_branch
git  push origin develop

PR #763, opened 2018-09-24
Add some tests

btry:
Add some tests

### Checklist

Please check if your PR fulfills the following specifications:

- [x] Tests for the changes have been added
- [ ] Docs have been added/updated

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@btry |1|

### References

Closes #N/A
Related #N/A
Depends on #N/A

PR #775, opened 2018-10-05
Update matrix for Travis CI

btry:
Signed-off-by: Thierry Bugier <tbugier@teclib.com>

### Changes description

PR #763 broke due to a recent change in GLPI 9.3/bugfixes, which impacts Fusion Ivnentory.

PR for Fusion Ivnentory : https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2590/files

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@btry |1|

### References

Closes #N/A
Related #N/A
Depends on #N/A


btry:
The failure was expected: the PR for Fusion Inventory is not merged yet. This PR will help to solve the issue by using the bugfixes branch for GLPI 9.3 bugfixes branch

btry:
> BTW we should be also a bit more consistent with same use for each case?, for example GLPI 9.2/bugfixes with FI branch glpi9.2

As GLPI 9.2 is not maintained anymore, I think we can simplify our matrix and test only the latest release of GLPI 9.2. Isers should upgrade to the latest reelase of GLPI 9.2 and consider quickly an upgrade to GLPI 9.3.

For Flyve, as we are late to release 2.0.0, We may consider to drop support of GLPI 9.2 and focus on GLPI 9.3. GLPI 9.3.2 is on its way already.

btry:
> BTW we should be also a bit more consistent with same use for each case?, for example GLPI 9.2/bugfixes with FI branch glpi9.2

An additional note: if such case occurs later, I strongly suggest to user a stable version of Fusion Inventory : changes in this project breaks its unit tests or our unit tests too often. we have frequently to workaround to make our tests green.

PR #789, opened 2018-10-29
Avoid QR code format change

### LINKED ISSUE #781: Image suffix of invitation email is .png but contains a JPG file
**Describe the bug**
the innovation mail "[GLPI] You have been invited to join Flyve MDM" contains an image as attachment. The filename is somehash.png. When viewing the image the type is JPEG, quality: 90, subsampling OFF

**To Reproduce**
1. Send innovation mail in glpi
2. Open attachment mail and check the type

**Expected behavior**
either the file type should be changed to PNG or the file suffix should be changed to .jpg so it matches its content.

I'm currently using Flyve Mobile Device Management 2.0.0-rc.2 and GLPI 9.3.1.

DIOHz0r:
### Changes description

GLPI resize the attached file if the image is less than 400px x 300px and convert it to JPEG format, this PR fixes the issue.

### Checklist

### Estimated time

|Assignee|:tomato:|
|:---|:---:|
|@DIOHz0r |6|

### References

Closes #781

### DIFF
--- a/inc/invitation.class.php
+++ b/inc/invitation.class.php
@@ -114,6 +114,13 @@ class PluginFlyvemdmInvitation extends CommonDBTM {
          return false;
       }
 
+      // does the user have a default email?
+      if (!$user->getDefaultEmail()) {
+         Session::addMessageAfterRedirect(__("Email address is invalid", 'flyvemdm'),
+            false, INFO, true);
+         return false;
+      }
+
       // The user already exists, add him in the entity
       $profile_User = new Profile_User();
       $entities = $profile_User->getEntitiesForProfileByUser($userId, $guestProfileId);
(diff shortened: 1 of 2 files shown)


inc/invitation.class.php
```diff
@@ -114,6 +114,13 @@ public function prepareInputForAdd($input) {
          return false;
       }
 
+      // does the user have a default email?
+      if (!$user->getDefaultEmail()) {
```
DIOHz0r: the list of users for adding an invitation shows all users without checking if their email is assigned

btry: hi,

don't fix this here : there is a workflow change in progress.

btry: I checked: the workflow change was merged. We might do better by filtering users in the dropdown when we create an invitation, but for now, I accept this.

PR #903, opened 2019-09-25
fix(core): fix wizard step

stonebuzz:
### Changes description

this pr solves the problem of the wizard that remains in step 1,
and also the number of the step

### Checklist

### References

Closes #N/A
Related #N/A
Depends on #N/A

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -56,14 +56,14 @@ class PluginFlyvemdmConfig extends CommonDBTM {
    const WIZARD_WELCOME_END = 1;
 
    // first and last steps of requirement pages of wizard
-   const WIZARD_REQUIREMENT_BEGIN = 100;
-   const WIZARD_REQUIREMENT_END = 107;
+   const WIZARD_REQUIREMENT_BEGIN = 2;
+   const WIZARD_REQUIREMENT_END = 9;
 
    // first and last steps of the MQTT pages of wizard
-   const WIZARD_MQTT_BEGIN = 108;
-   const WIZARD_MQTT_END = 111;
+   const WIZARD_MQTT_BEGIN = 10;
+   const WIZARD_MQTT_END = 13;
 
-   const WIZARD_FINISH = -1;
+   const WIZARD_FINISH = 14;
    static $config = [];
 
    /**
@@ -364,6 +364,7 @@ class PluginFlyvemdmConfig extends CommonDBTM {
       if (!isset($_SESSION['plugin_flyvemdm_wizard_step'])) {
          $_SESSION['plugin_flyvemdm_wizard_step'] = static::WIZARD_WELCOME_BEGIN;
       }
+
       echo '<form name="form" id="pluginFlyvemdm-config" method="post" action="' . Toolbox::getItemTypeFormURL(__CLASS__) . '">';
 
       $texts = [];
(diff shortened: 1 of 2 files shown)


inc/config.class.php
```diff
@@ -421,12 +423,10 @@ public static function configUpdate($input) {
...
 
       if (isset($_SESSION['plugin_flyvemdm_wizard_step'])) {
-         $input = static::processStep($input);
-         if (count($input) > 0 && $input !== false) {
-            static::forwardStep($input);
-         } else {
-            $input = [];
-         }
+         //disable wizard if needed
+         static::processStep($input);
```
btry: if you drop the argument in the method, you no longer need to pass $input here.

Initially, passing the $input is to allow some actions based on the data filled in a form embedded in the step. This would be nice to keep this ability.

inc/config.class.php
```diff
@@ -56,14 +56,14 @@ class PluginFlyvemdmConfig extends CommonDBTM {
...
+   const WIZARD_REQUIREMENT_END = 9;
 
    // first and last steps of the MQTT pages of wizard
-   const WIZARD_MQTT_BEGIN = 108;
-   const WIZARD_MQTT_END = 111;
+   const WIZARD_MQTT_BEGIN = 10;
+   const WIZARD_MQTT_END = 13;
 
-   const WIZARD_FINISH = -1;
+   const WIZARD_FINISH = 14;
```
btry: Numbers were chosen to make easier to add pages in the wizard without changing all constants. If we add a page in the MQTT part for example, we don't need to change all parts occuring after it.

WIZARD_FINISH = -1 to also not bother about this value when we change the steps count.

tpl/config-wizard.html.twig
```diff
@@ -281,7 +281,7 @@ ciphers ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES2
 		</tr>
 		<tr class="tab_bg_1">
 			<td class="right" colspan="2">
-			{{ __('Wizard page:','flyvemdm') }}&nbsp;{{ step }}
+			{{ __('Wizard page:','flyvemdm') }}&nbsp;{{ step }} / {{total_step}}
```
btry: This number is a technical information to more easily debug the wizard.

stonebuzz: I agree with you, but going from step 1 to step 100 is confusing for users

inc/config.class.php
stonebuzz: I agree with you, but going from step 1 to step 100 is confusing for users

inc/config.class.php
stonebuzz: fixed

inc/config.class.php
stonebuzz: how do you imagine this part

tpl/config-wizard.html.twig
stonebuzz: how do you imagine this part

btry: We may just change the label to tell this is a technical ID ?

stonebuzz: Done

stonebuzz: Wizard page is renamed to Technical step ID

